Issue #3, opened 2016-12-08
Investigate if we can avoid multiple instances of the DataManager


iocanel:
Currently, we use a data manager instance per entity.

Maybe we could create a single instance of the datamanager and inject it wherever we need it using cdi.

PR #6, opened 2016-12-08
#3 DataManager is now managed by CDI.

iocanel:
The benefit is that we now have a single isntance of the DataManger which is initialized just once.

@KurtStam: Not sure If this breaks something, so wdyt?


KurtStam:
Works great

PR #12, opened 2016-12-09
Working swagger ui custom index.html and pom updates

### LINKED ISSUE #1: swagger-ui does not use custom index.html when running in container
Can we upgrade the openjdk version to 1.8.0_111 in the s2i container?

jimmidyson:
Fixes #1


runtime/pom.xml
```diff
@@ -73,6 +72,7 @@
             <goals>
               <goal>resource</goal>
               <goal>helm</goal>
+              <goal>build</goal>
```
iocanel: This breaks the build on circleci.

iocanel: Hmmm, I thought that it was already the case. Let me see.

jimmidyson: It's foobarred - no hook gets sent it seems. Enabling the project from circleci doesn't actually add the webhook either...

PR #19, opened 2016-12-11
Add generation of static HTML docs served from API server

jimmidyson:
When run with `java -jar target/swarm-runtime.jar`, static HTML docs are served
from http://localhost:8080. Haven't figured out how to get those embedded in dev
mode, but that's less important anyway.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,17 @@
     </repository>
   </distributionManagement>
 
+  <repositories>
+    <repository>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>jcenter-releases</id>
+      <name>jcenter</name>
+      <url>https://jcenter.bintray.com</url>
+    </repository>
+  </repositories>
+
   <pluginRepositories>
     <pluginRepository>
       <snapshots>
@@ -80,7 +91,18 @@
       </snapshots>
       <id>jcenter-releases</id>
       <name>jcenter</name>
-      <url>http://jcenter.bintray.com</url>
+      <url>https://jcenter.bintray.com</url>
+    </pluginRepository>
+    <pluginRepository>
+      <id>apache.snapshots</id>
+      <name>Apache Development Snapshot Repository</name>
+      <url>https://repository.apache.org/content/repositories/snapshots/</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
     </pluginRepository>
   </pluginRepositories>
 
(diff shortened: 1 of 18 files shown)


jimmidyson:
@kahboom Hope that this should produce some static HTML docs for the REST API to make it easier for you to review them as needed & mean we don't need to manually update them as the implementation progresses. @KurtStam it does mean we need to keep the swagger annotations up to date & verbose though :)

jimmidyson:
Had to use a SNAPSHOT dependency plugin to get circle to pass (old version is not Maven 3 compatible it seems). New version should be released soon.

jimmidyson:
See example static doc output at https://32-75404146-gh.circle-artifacts.com/0/tmp/circle-artifacts.W198HOB/apidocs.html. We can customize templates as we need but we can sort that out later,

runtime/src/main/java/com/redhat/ipaas/rest/Connections.java
```diff
@@ -30,58 +33,61 @@
...
-			@ApiParam(value = "id of the connection", required = true) @PathParam("id") String id) {
-		dataMgr.delete(Connection.class, id);
-		
-	}
-	
-	
-	
-}
\ No newline at end of file
+    @Inject
```
rhuss: We should settle on a convention of the indent in Java source files. I'd prefer 4 spaces overall, (not 2 for the initial one) but would anything we agree upon. But we should stick to the same convention to not mess up diffs.

jimmidyson: Agreed 4 spaces for Java.

Sorry about the formatting changes - got any tips on something to check style (checkstyle I assume?) that works well & is not too intrusive? P.S. I love gofmt :)

jimmidyson: I've set that up in `.editorconfig` already (https://github.com/redhat-ipaas/ipaas-api-java/blob/master/.editorconfig#L8-L9). See http://editorconfig.org if you don't know about editorconfig (it's got pretty wide editor support via plugins & is nice & simple).

jimmidyson: Yeah I use intellij too. Raised #20 to discuss more.

pom.xml
```diff
@@ -100,6 +122,7 @@
     <maven.resources.plugin.version>3.0.1</maven.resources.plugin.version>
     <maven.source.plugin.version>2.4</maven.source.plugin.version>
     <maven.surefire.plugin.version>2.19.1</maven.surefire.plugin.version>
+    <maven.dependency.plugin.version>3.0.0-SNAPSHOT</maven.dependency.plugin.version>
```
iocanel: Do we even use the maven dependency plugin?
And why do we need a snapshot version.

I am sure there is a reason, just being curious.

jimmidyson: For circleci to speed up caching dependencies between builds `mvn dependency:go-offline` happens in dependency stage before cache is created for next build, but before actual build is run (don't want to cache build artifacts from this build).

jimmidyson:
Anything need fixing up? Please can someone approve the PR & merge it if you are or tell me what needs changing if not. Pretty please?

KurtStam:
Sure - taking a look now.

KurtStam:
[merge]

jimmidyson:
Sorry no CI for this yet on Jenkins - I can set that up tonight.

PR #21, opened 2016-12-12
Add code coverage using codecov.io

jimmidyson:


jimmidyson:
@KurtStam You can see it on codecov - e.g. https://codecov.io/gh/redhat-ipaas/ipaas-api-java/pull/21?src=pr. The sunburst chart is interactive too btw,

PR #25, opened 2016-12-12
Use convenient proxy for Swagger badge to work around https://github.com/swagger-api/validator-badge/issues/77

jimmidyson:


jimmidyson:
Just a minor readme update so self merging

PR #27, opened 2016-12-13
adding CRUD for tags

KurtStam:


api/src/main/java/com/redhat/ipaas/api/IPaasEntity.java
```diff
@@ -15,9 +15,15 @@
...
-public interface IPaasEntity {
+public abstract class IPaasEntity {
 
-	public String getId();
-	public void setId(String id);
+	abstract public String getId();
+	abstract public void setId(String id);
+	
+	@Override
+	public boolean equals(Object obj) {
```
jimmidyson: Why do we need to override equals just for ID? Can't we just compare IDs when required or use a custom Comparator to do that?

jimmidyson:
ok to test

api/src/main/java/com/redhat/ipaas/api/IPaasEntity.java
KurtStam: I'm expressing that two IPaasEntities are equal when their IDs match. Makes it easier to work with them in Lists too and just less error prone in general as they are the same object when their IDs match..

jimmidyson: And what happens to `hashCode`?

jimmidyson: Seeing as we're using java 8, I'd prefer to use something like `list.stream().anyMatch(new IdPredicate())` (a utility function providing that).

jimmidyson: Because they're not equal just because of ID - you're just checking that you don't have items with duplicate IDs in the list. That is not object equality. This will hurt us if we ever want to use other data structures, e.g. Sets, that contain these objects.

Plus what is more readable about `list.contains(obj)` than `containsId(list, obj)` where `containsId` is a utility function to do the check. To my mind that's arguably _more_ readable.

KurtStam: The Ids are used a PK in the cache. We should not allow having multiple objects with the same ID(PK) in there. I think it's perfectly ok to override equals() to support this.

jimmidyson: There are loads of other things that you want to check for before you put stuff in the cache. ID is just one case. How about e.g. name? What you're looking for is uniqueness on a (number of) field(s) in the cache (ala unique key in a DB table).

api/src/main/java/com/redhat/ipaas/api/IPaasEntity.java
```diff
@@ -15,9 +15,22 @@
...
```
jimmidyson: It also makes testing harder as we can't rely on equals to test deep equality so will have to test individual field equality which sucks.

runtime/src/main/java/com/redhat/ipaas/rest/Connections.java
```diff
@@ -85,8 +93,50 @@ public void update(
...
         @ApiParam(value = "id of the connection", required = true) @PathParam("id") String id) {
         dataMgr.delete(Connection.class, id);
-
     }
-
-
+    
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path(value = "/{id}/collection/tags")
```
jimmidyson: Why do we need the `/collection/` segment?

KurtStam: I'm following the restful objects specification: http://www.restfulobjects.org.

jimmidyson: That seems to be a dead spec (last updated 2 years ago) & doesn't seem to have caught on. Most real world REST APIs don't have this segment which seems redundant & doesn't fulfil any purpose other than making the URL longer. `/{id}/tags` is enough & is in much more common usage, a de facto standard I would say.

KurtStam:
There you go, anything else?

jimmidyson:
This is going to need a big ol' rebase after #34 was merged - I don't mind doing it if you give me push karma to your repo.

KurtStam:
@jimmidyson np, won't take me that long.

KurtStam:
It'd be nice to get a review on this. It's been sitting here for a while now.

rhuss:
@KurtStam I could do a review later, but think @jimmidyson is probably better suited.

jimmidyson:
Sorry I haven't looked at this but I've had a lot of other things to do as you might have noticed. Right now I'm not too interested in adding functionality to the REST API while the model is still in flux, as we're discussing in the ipaas brain storming doc.

I would prefer to see stuff that we know we're going to have to do like authentication, authorization, logging, tracing, instrumentation, etc while we figure out what the model & finally the REST API should look like.

How does that sound?

KurtStam:
It's something Rachel asked for specifically for he demo.

kahboom:
Thanks @KurtStam - I had asked around and Tags were still a model that would be necessary. Not sure about the priority though, @jimmidyson has a better idea of that.

Issue #10, opened 2016-12-08
Add Application Configuration


KurtStam:
For things like:

- Running with or without back-end (in this case preload data from file)
- Configuration of the back-end endpoint access
- Authentication and Authorization Store
- Activate Session Recording (to collect data to run w/o backend)

iocanel:
What do you have in mind for application configuration? 
I guess sicne we are using swarm and cdi, deltaspike may be the obvious choice. Thoughts?

KurtStam:
Go for it :)

jimmidyson:
Learning wildfly swarm slowly... How about https://wildfly-swarm.gitbooks.io/wildfly-swarm-users-guide/content/configuration/project_stages.html?

iocanel:
Project stages looks good to me and it seems that it works with CDI too, so let's use that.

jimmidyson:
Done now we've switched to spring boot.

PR #29, opened 2016-12-13
#10 Add project-stages.yml. Replace hardcoded value in DataManager wi…

iocanel:
…th @Inject/@ConfigurationValue from project-stages.yml.


iocanel:
retest this please

Issue #22, opened 2016-12-12
Add generic list sorting feature


KurtStam:
The UI requires a feature to sort and filter lists. I think this maybe on searches. It'd make sense to come up with generic search and sort that we can hook into the REST api.

jimmidyson:
This should be on all list requests (`GET` on resources, not a separate search operation), fields that are filterable/searchable will be dependent on the type being requested.

jimmidyson:
Closing now - sorting is fixed #32.

PR #32, opened 2016-12-13
Add support for sorting for all list operations (#22)

rhuss:
You can use the following query parameters:

* "sortField" to specify on which field to support. Supported are currently integer and string fields.
* "sortDir" can be "asc" or "desc" dependening on which direction to sort. Default is "asc"

Still open: Proper Swagger annotation which is probably done via a "Parameters Definition Object" (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameters-definitions-object) so that it doesnt not have to repeated all over all.

Also it would be nice to sort on multiple levels (by providing multiple sort fields) and also on deeper values by allowing a path expression for the field. But thats probably over the top for now ;-)

Interesting also: The DataManager.fetch(...) method can take a list of transformers (like the ReflectiveSorter), so it should be trivial to add paging at that point, too.


runtime/src/main/java/com/redhat/ipaas/rest/ComponentGroups.java
```diff
@@ -37,12 +38,15 @@
...
+    @Context
+    private UriInfo uri;
+
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "List component groups")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "Success", response = ComponentGroup.class)})
     public Collection<ComponentGroup> list() {
-        return dataMgr.fetchAll(ComponentGroup.class);
+        return dataMgr.fetchAll(ComponentGroup.class, new ReflectiveSorter<>(ComponentGroup.class, uri.getQueryParameters()));
```
jimmidyson: Does reflective sorter need all query params? Can we just pass in field and direction?

rhuss: Yes, we could. But I thought it would better to let the sorter pick the properties he wants to, so that the endpoints don't kneed the knowledge. Also in case of a change or maybe an additional parrameter to sort (e.g. extracting multiple fields for a secondary sort order), one would only need to change the sorter. 

The drawback of course is that it couples the ReflectiveSorter to jax-rs' MultiValueMap so that it cant be reused in a different context.

jimmidyson: Yeah I'd rather separate that out. How about an intermediate `SortOptions` to pass to the `ReflectiveSorter` that's created from the context?

rhuss: But then again you would either build up the `SortOptions` by specifying both parameters `sortField` and `sortDir` as parameters so that each fetchAll() method would need to extract it by themselve or you would couple `SortOptions` also to `MultiValueMap`.

The only solution I would see it to copy over the whole MultiValueMap into a regular Map (but the losing multi-values which might help when introducing multiple sortFields for multiple sort order) or put it behind a facade (which might be the simpler solution). Let me try the latter.

jimmidyson: It's the responsibility of the rest service to take the incoming raw data (body, query string, cookies, etc) & turn it into a representation that internal services that it calls can use, in this case it's translating the query string parameters for sort into a form that the sorter can use, SortOptions. Why is it a problem to do this in every list operation, with perhaps a method in an introduced superclass so we can get some more consistency or via a simple extractor utility?

rhuss: Gimme a second, I think I have a good solution.

rhuss: Updated the PR. I could now move the sorter also to the API module as it doesn't has any refs to jax-rs anymore. Does this make sense ?

jimmidyson:
LGTM. Will take full look when I'm on my computer later.

runtime/src/main/java/com/redhat/ipaas/rest/SortOptionsFromQueryParams.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+class SortOptionsFromQueryParams implements SortOptions {
+    private final MultivaluedMap<String, String> queryParams;
+
+    private static final String SORT_FIELD = "sortField";
+    private static final String SORT_DIRECTION = "sortDir";
+
+
+    SortOptionsFromQueryParams(UriInfo uri) {
+        queryParams = uri.getQueryParameters();
```
jimmidyson: Just extract the sort field & dir in the constructor. No need for calculating in each getter.

rhuss:
@jimmidyson If you don't mind (and if its ok now), I would squash the commits now ....

jimmidyson:
Overall it looks OK for now but there are a number of questions this raises for future use of the cache. Please squash & we can merge this in, fix up outstanding stuff later.

Not a big fan of reflection at runtime (performance - although) - wonder if we can do some generation here based on annotations for sortable fields? @iocanel?

@KurtStam Do we need to add some indexing into the cache so we can ask the cache for sorted results? We'll need the same when we add filtering too, it's just going to get too expensive to list all & then sort/filter in a way similar to this PR.

Also, how do we handle default sort order? Relying on cache insertion order isn't going to give a very consistent experience.

jimmidyson:
@rhuss This is failing license check still btw.

rhuss:
w.r.t to caching, agreed that there are many open questions. Also in case we are using a real backend for storing stuff where sorting could be done there, too.

w.r.t to reflection: Agreed that its not as good as generated accessors, but it is simple and so I would like to defer optimisation to a later point, when needed. No prematured optimisation, at least we know where to look if things are slow w.r.t to sorted lists.

w.r.t to default sort oder: For me a natural ordering would be according to the id (if its monotonically increasing as it should be) or insertion date (if not). An integer id would also make comparing easier (currently it would be compared a strings which is not good when not having a fixed number of digits for the ids).

jimmidyson:
As discussed with @rhuss, let's switch to follow github's API style & use `sort` & `direction` for query parameters.

jimmidyson:
> No prematured optimisation

Agreed, but it will be necessary, I can feel it :)

> For me a natural ordering would be according to the id

Feels more like name to me, as ID is hidden in UI so to the user, the order would look pretty random.

rhuss:
Sorting by id, if monotonically increasing is a nice idea to get the insertion order, or sorting 'by age'. Even when its not visible.

rhuss:
btw changed to 'sort' and 'direction', too and squashed all to a single commit.

runtime/src/main/java/com/redhat/ipaas/rest/Components.java
```diff
@@ -38,12 +35,17 @@
...
     private DataManager dataMgr;
 
+    @Context
+    private UriInfo uri;
+
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @ApiOperation(value = "List components")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "Success", response = Component.class)})
+    @ApiParam(name = "sortField", value = "Sort list with the given field")
```
rhuss: Ah, sorry. Need to be removed as this doesn't work (was just a test). Swagger doesnt pick this up. 

I need to find out to add this externally as described in https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameters-definitions-object

jimmidyson:
Need to rebase after #34 was merged - I don't mind doing that now if you're OK with that & I'll have to force push to your branch (able to do so now thanks to [this](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/)),

rhuss:
fine for me, go ahead.

jimmidyson:
Turns out I can't push to your fork's branch :( I've rebased & works great so can you give me push karma on your repo when you get a chance (can wait until tomorrow of course!) & it will be done.

rhuss:
you have karma now

PR #34, opened 2016-12-13
Immutable model resources

jimmidyson:
This makes a change to immutable model resources for all the good immutability reasons. Uses [Immutables](https://immutables.github.io) to generate immutable package private implementations from the annotated interface definitions. I think it makes the model more readable & removes quite a bit of repetitive hand-written code to maintain.

I've tested this code with the iPaaS client & all works great, without changes to frontend. I had to add a CORS filter (albeit totally open for now) & customised Jackson `ObjectMapper` so kept this in this PR.


jimmidyson:
Sorry for the messy diff, git didn't like the move & edit for some reason, hence it looks like deletes & creates :(

jimmidyson:
Seeing as this quite an invasive change I'd appreciate a review (on the approach as much as the code) so I don't have to keep rebasing as other changes are made.

jimmidyson:
Yeah sorry about the bad diffs :( Why do we need @JsonRaw? Isn't it just embedded JSON as strings?

jimmidyson:
When I look at the example https://github.com/redhat-ipaas/ipaas-api-java/blob/master/api/src/main/resources/com/redhat/ipaas/rest/deployment.json#L379 it looks like it should be an escaped string that @kahboom & @gashcrumb are expecting in the response. \@JsonRawValue won't escape the string like that.

KurtStam:
They told me that they wanted a string containing a JSON blob.

jimmidyson:
Right so you don't want JsonRawValue then, unless I'm misunderstanding what https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonRawValue.html explains: `the annotated method or field should be serialized by including literal String value of the property as is, without quoting of characters`.

KurtStam:
So can we still use (JAX-RS) annotations on getters/setters if we need to yes or no? You would do that by add your won getter/setter if you have to?

jimmidyson:
Well there are no setters which would break immutability, but yes you can annotate the accessors.

jimmidyson:
Noticed dropping the get prefix messes up generated swagger. Will fix that up soon but please don't merge. We should add some validation of swagger spec to ensure it has everything we need, not just that it's valid.

jimmidyson:
Switched to get prefixes & swagger spec looks good again. Will merge on green.

jimmidyson:
[merge]

PR #35, opened 2016-12-13
Put fabric8 mvn tooling in fabric8 profile

### LINKED ISSUE #30: Support running outside of kubernetes
For what reason do we involve fmp during the build? I like running

mvn clean install

from the parent module to compile the whole thing but that requires a running k8s. Do we really need that?

jimmidyson:
Fixes #30

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -79,16 +79,6 @@
         <groupId>io.fabric8</groupId>
         <artifactId>fabric8-maven-plugin</artifactId>
         <version>${fabric8.maven.plugin.version}</version>
-        <executions>
-          <execution>
-            <id>fmp</id>
-            <goals>
-              <goal>resource</goal>
-              <goal>helm</goal>
-              <goal>build</goal>
-            </goals>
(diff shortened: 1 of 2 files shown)


runtime/pom.xml
```diff
@@ -250,4 +237,29 @@
...
+    <profile>
+      <id>fabric8</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>fabric8-maven-plugin</artifactId>
+            <version>${fabric8.maven.plugin.version}</version>
+            <executions>
+              <execution>
```
rhuss: I would leave the configuration outside the profile and only remove the executions or move the executions into a profile. That way people can use goals like 'fabric8:resource' without specifying this profile.

jimmidyson: Done. Added executions in `fabric8` profile.

rhuss:
thanks !

lgtm

rhuss:
[merge]

Issue #24, opened 2016-12-12
Add pagination to list requests


jimmidyson:
All requests for resource lists (e.g. GET resources) should accept pagination parameters in the query string. Responses should contain generic details on the list, e.g. total count.

jimmidyson:
Fixed in #58

PR #37, opened 2016-12-15
Add support for sorting for all list operations

rhuss:
You can use the following query parameters:

* "sort" to specify on which field to support. Supported are currently integer and string fields.
* "direction" can be "asc" or "desc" dependening on which direction to sort. Default is "asc"

Still open: Proper Swagger annotation which is probably done via a "Parameters Definition Object" (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameters-definitions-object) so that it doesnt not have to repeated all over all.

Also it would be nice to sort on multiple levels (by providing multiple sort fields) and also on deeper values by allowing a path expression for the field. But thats probably over the top for now ;-)

Interesting also: The DataManager.fetch(...) method can take a list of transformers (like the ReflectiveSorter), so it should be trivial to add paging at that point, too.

### DIFF
--- a/runtime/src/main/java/com/redhat/ipaas/rest/Users.java
+++ b/runtime/src/main/java/com/redhat/ipaas/rest/Users.java
@@ -17,11 +17,7 @@ package com.redhat.ipaas.rest;
 
 import com.redhat.ipaas.api.v1.model.User;
 import com.redhat.ipaas.rest.util.ReflectiveSorter;
-import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiOperation;
-import io.swagger.annotations.ApiParam;
-import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
+import io.swagger.annotations.*;
 
 import javax.inject.Inject;
 import javax.ws.rs.GET;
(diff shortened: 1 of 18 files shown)


runtime/src/main/java/com/redhat/ipaas/rest/Users.java
```diff
@@ -37,12 +40,16 @@
     @Inject
     private DataManager dataMgr;
 
+    @Context
```
jimmidyson: This is just a matter of personal taste, but I prefer this to be a parameter to the method, just feels a bit more readable to me for request scoped context. What do you think?

rhuss: yeah, thats probably better as its more direct. tbh, I always have to think how this is injected on a per request basis. Always like to think of endpoints as immutable singletons, but thats not the reality. 

The original motivation here was to keep method signature clean, not cluttering it with 'technical' details. 

will change it.

rhuss:
@jimmidyson I'm done so far. If this PR is approved and merged I would continue on #24 .

jimmidyson:
LGTM thanks! [merge]

PR #39, opened 2016-12-15
WIP: Composable REST API

jimmidyson:
Would love feedback on this approach to ensure a consistent API with little repeated code. The approach uses default methods on interfaces to act like traits to compose functionality in the REST API layer.

One outstanding problems is Swagger descriptions: generics don't play nicely with swagger and description strings in annotations need to be constants. I'm thinking of using annotation processing to generate implementations of these interfaces, supplying annotations on the managed model classes. These generated classes would of course not be editable, but we can also generate a subclassed implementation where we can extend or override as needed.

I've tested this with the iPaaS UI & it all works fine :)

### DIFF
--- a/runtime/src/main/java/com/redhat/ipaas/rest/BaseHandler.java
+++ b/runtime/src/main/java/com/redhat/ipaas/rest/BaseHandler.java
@@ -15,23 +15,16 @@
  */
 package com.redhat.ipaas.rest;
 
-import com.redhat.ipaas.api.v1.model.User;
-import io.swagger.annotations.Api;
+import javax.inject.Inject;
 
-import javax.ws.rs.Path;
+public class BaseHandler implements WithDataManager {
 
-@Path("/users")
-@Api(value = "users")
-public class Users extends BaseHandler implements Lister<User>, Getter<User> {
+    @Inject
+    private DataManager dataMgr;
 
     @Override
-    public Class<User> resourceClass() {
-        return User.class;
-    }
-
-    @Override
-    public String resourceKind() {
-        return User.KIND;
+    public DataManager getDataManager() {
+        return dataMgr;
     }
 
 }
--- a/runtime/src/main/java/com/redhat/ipaas/rest/ComponentGroups.java
+++ b/runtime/src/main/java/com/redhat/ipaas/rest/ComponentGroups.java
@@ -16,49 +16,22 @@
 package com.redhat.ipaas.rest;
 
 import com.redhat.ipaas.api.v1.model.ComponentGroup;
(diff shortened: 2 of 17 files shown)


jimmidyson:
This includes the changes from #37 which I hope we can merge soon.

runtime/src/main/java/com/redhat/ipaas/rest/BaseHandler.java
```diff
@@ -0,0 +1,15 @@
...
+
+import javax.inject.Inject;
+
+public class BaseHandler implements WithDataManager {
+
+    @Inject
+    private DataManager dataMgr;
+
+    @Override
+    public DataManager dataManager() {
```
iocanel: Why you denounce getters?

iocanel:
Looks brilliant.

runtime/src/main/java/com/redhat/ipaas/rest/BaseHandler.java
jimmidyson: They just feel so unnecessary... Death to useless keystrokes.

jimmidyson:
Reaction overload from @kahboom.

runtime/src/main/java/com/redhat/ipaas/rest/BaseHandler.java
rhuss: fwiw for the model objects the ReflectiveSorter relies on having get-methods (but could be adapted if one feels so)

jimmidyson: No that's fine - also the swagger generation requires getters, but that doesn't mean you have to use them when you don't need to. Or maybe I'm just being too radical, been corrupted by too many other languages now.

iocanel: There are tons of tools and frameworks that make use of the getter convention and so little gain.

jimmidyson: Fair enough. But let it be known that I hate it.

jimmidyson: Done :sob:

jimmidyson:
@KurtStam I'm all about over-engineering :-b Seriously though I do agree for this small REST API it may look like being over-engineered but I wanted to try out a technique to introduce consistency at a layer where there is normally a lot of repetitive code that can easily lose that consistency without strict review or generation. Even in the infant state of this API, we have different functional support for resources (get, list, delete, update, etc) that should have been consistent. And, yes, there's less code to maintain.

PR #41, opened 2016-12-15
Adopt Google Java format as code style

### LINKED ISSUE #20: Code style
Re discussion in https://github.com/redhat-ipaas/ipaas-api-java/pull/19#discussion_r91928245 we should settle on a code style that we can all adopt to ensure that PRs are easier to review (formatting changes not muddying the changes for review).

We have to remember that not everyone uses the same editor (although of course IntelliJ is the best for Java ;)).

And we probably need to have an automated check for PRs so that we can enforce the code style.

jimmidyson:
Fixes #20

### DIFF
--- a/api/src/main/java/com/redhat/ipaas/api/ImmutablesStyle.java
+++ b/api/src/main/java/com/redhat/ipaas/api/ImmutablesStyle.java
@@ -18,14 +18,13 @@ package com.redhat.ipaas.api;
 import org.immutables.value.Value;
 
 @Value.Style(
-    jdkOnly = true,
-    visibility = Value.Style.ImplementationVisibility.PACKAGE,
-    defaultAsDefault = true,
-    headerComments = true,
-    depluralize = true,
-    typeAbstract = "*",
-    allParameters = true,
-    from = "createFrom"
+  jdkOnly = true,
+  visibility = Value.Style.ImplementationVisibility.PACKAGE,
+  defaultAsDefault = true,
+  headerComments = true,
+  depluralize = true,
+  typeAbstract = "*",
+  allParameters = true,
+  from = "createFrom"
 )
-public @interface ImmutablesStyle {
-}
+public @interface ImmutablesStyle {}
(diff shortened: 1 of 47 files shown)


rhuss:
If the review is about an opinion on the style chosen, than its ok for me, except the two chars indentation instead of four, which I never seen before in other projects tbh. Really would love to stick to 4, it much easier to read for me ....

Regardless what we do, we should decide quickly on this PR because of potential conflicts when we would have to rebase other WIP.

rhuss:
Jenkins, you are talking with us ?

jimmidyson:
Must be Jenkins 2 :-)

The reason for adopting a code style here is really to reduce noise in PR changes for review (but then you might not like PR reviews either if my guess is right why you really are :-b). I'm not even fussed what the style is, just that it is consistent & easy to adopt & validate via tooling to do formatting rather than manually.

rhuss:
I agree that the golang style to enforce a formatting as part of the  language is a good thing w.r.t collaboration, although it feels quite uncomfortable for Java engineers at first. 

The problems with cluttered PRs could be also resolved if everyone would resists to adapt the existing formatting (even when it might be inconsistent). I know for myself that it's hard to resist to add a whitespace there and there, and I can't guarantee to stop this as it is a reflex :).  Maybe a compromise could be to automatic reformatting periodically and in between ppl can work freely. I guess that in the mid-term everybody will automatically adapt to the given style as they become tired.

But I agree with Jenkins 2 that we should not spend too much effort on this. As said, I'm not really biased (except for the spaces ;-P

iocanel:
I think that using a maven plugin is a good idea.
I honestly hate the 2 space indentation for java.

jimmidyson:
@iocanel Do you know of any other maven plugin that can do the formatting? Sounds like you @rhuss hate two spaces enough that we shouldn't adopt this style at least.

iocanel:
@jimmidyson: isn't that configurable?

iocanel:
If not we need to find one that respects .editorconfig.

And if we can't find one, we should ask our friends at dev experience to create one for us.

jimmidyson:
It is not configurable, that's one of the nice things about it to my mind, opinionated to reduce arguments like this :-b

We could just fork it & set indentation to 4 of course.

chirino:
I'm all for this as long as the auto format feature in a default Intelij automatically does the right thing :)

PR #43, opened 2016-12-16
Run circleci tests in docker container to work around old version of maven

jimmidyson:


jimmidyson:
Just going to [merge] this as it is only for circleci testing

PR #44, opened 2016-12-16
Move v1 REST endpoints to api module to separate runtime

jimmidyson:
Plus some other minor changes to dependencies and test utils

Main purpose of this is to separate out our REST API from our swarm packaging in case we want to ditch swarm later and move to spring boot.


KurtStam:
LGTM - I have revolved the conflicts in PR #52

PR #46, opened 2016-12-16
Deploy to openshift dedicated from circleci on every push to master

jimmidyson:


jimmidyson:
Going to [merge] this so I can try it out... here goes. Will revert if it messes anything up,

Issue #28, opened 2016-12-13
Add ExceptionManager to REST calls


KurtStam:
Add an [exception mapper](https://docs.jboss.org/resteasy/docs/2.2.1.GA/userguide/html/ExceptionHandling.html) and return appropriate status codes on exceptions.

PR #47, opened 2017-01-03
Add ExceptionManager to REST calls #28

KurtStam:


iocanel:
I am not a big fan of checked exceptions, mostly do to the noise they add...

So, I would like to propose create a new unchecked exception say, IPaasException that would be used to wrap around all kinds of checked exceptions. This would remove the noise from methods, and would still allow us to use something like an ExceptionMapper. 

I think that this concept is suggested in `Effective Java` and we have been using it in the Kuberntes Client too: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClientException.java#L51

Now in this particular case, we could store the root exception so that it can be used in the mapper if needed. 

THoughts?

rhuss:
I think mostly the same, especially if we would to add mappers for every exception (so maybe a single REST exception handler which creates different errors based on the wrapped 'real' exception ?). Exception tunneling makes sense to me in general, especially when go across layers.

@iocanel btw, why don't you throw the runtime exception here in your example but return it ? Whereas for a wrapped checked exception you throw it ?

iocanel:
@rhuss: Currently the method is used like:

    catch (Throwable t) {
       throw KubenetesClientException.launderException(t);
    }

If we threw the exception instead of returning the method would be used like:

    catch (Throwable t) {
        KubenetesClientException.launderException(t);
    }

The first seems a little bit more clear. Everyone gets what it does, without having to look at its impl. Other than that no real reason.

iocanel:
@rhuss: If you are refering to line 57, then yeah I guess we should return the Exception there too :-)

rhuss:
@iocanel, ah yeah, sorry forgot the link ;-) But you got it anyway :)

jimmidyson:
:+1: to unchecked exceptions.

KurtStam:
The discussion is a bit besides the point as the exception mapper is used to map java exceptions to http status/error codes, with the exceptions being checked or unchecked, and then being able to providing the REST user with correct http status and and a useful message.

Let's open a new issue to change our error handling to unchecked exceptions.

iocanel:
That's fine with me.

PR #48, opened 2017-01-04
Issue11 hookup keycloak

### LINKED ISSUE #11: Hooking up KeyCloak
Hooking into KeyCloak for authentication, but also to lookup Users, Roles and Permissions.

KurtStam:
Fixes #11

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,3 +35,24 @@
 
 There is a [Demo endpoint](http://runtime-kurt.b6ff.rh-idev.openshiftapps.com/swagger-ui/) on OpenShift dedicated.
 This demo endpoint has some preloaded data and can be used for testing and demoing purposes.
+
+# Authentication
+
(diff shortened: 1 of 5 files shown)


KurtStam:
'Unpicking' involves removing the one keycloak-server reference in the pom. The rest of the configuration will remain all the same if keycloak is deployed elsewhere. Personally I think it makes the development process that much easier to have keycloak right in the server, so the UI developers only have to run one command to get the REST api they need. Please let's keep this as simple as possible to work on, until we know what keycloak instance we are really going to use.

jimmidyson:
> until we know what keycloak instance we are really going to use

I think we'll have to deploy our own - I expect that to be part of your investigation in this sprint into getting Keycloak hooked up properly.

But if it unblocks us for now then let's go with it & remove ASAP.

Issue #49, opened 2017-01-04
Switch to unchecked exceptions


iocanel:
As discussed here: https://github.com/redhat-ipaas/ipaas-api-java/pull/47 there is concensus to wrap checked exception inside an unchecked exception to reduce the noise is the source.

KurtStam:
@iocanel does that look ok to you? Or was there anything else had in mind?

PR #53, opened 2017-01-13
Switch to unchecked exceptions #49

KurtStam:


api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -101,20 +103,27 @@ public void init() {
...
+            Optional<String> id = entity.getId();
+            String idVal;
+            if (!id.isPresent()) {
+                idVal = generatePK(entityMap);
+                entity = entity.withId(idVal);
+            } else {
+                idVal = id.get();
+            }
+            entityMap.put(idVal, entity);
+        } catch (ClassNotFoundException e) {
```
iocanel: Since we don't handle them differently we could possibly collapse them into Exception. No biggie though!

api/src/main/java/com/redhat/ipaas/api/v1/rest/exception/IPaasServerException.java
```diff
@@ -0,0 +1,28 @@
+package com.redhat.ipaas.api.v1.rest.exception;
+
+public class IPaasServerException extends RuntimeException {
```
iocanel: I think that what's missing here is a method that take a throwable and either wrap it inside an `IPaasServerException` or re-throw it if it a runtimeexception (e.g. IllegalArgumentException or EntityNotFoundException).

KurtStam: Hi @iocanel I added the 'hot wash' laundry code. LMK if this looks ok.

KurtStam:
[merge]

PR #57, opened 2017-01-16
Use released maven dependency plugin

### LINKED ISSUE #55: maven-dependency-plugin version is set to 3.0.0-SNAPSHOT
Is there any reason to use snapshot version of dependency plugin?

https://github.com/redhat-ipaas/ipaas-api-java/blob/master/pom.xml#L131

jimmidyson:
Fixes #55


jimmidyson:
Self merging as no code changes.

PR #65, opened 2017-01-20
Integrations

iocanel:
This is work in progress, which means that you can review, provide feedback, but its not to be merged yet.

The end goal is to create a back-end for Integrations. 

At this point an integration can be a configmap or secret, with a special kind of label that indicates its an integration. The back end can be a simple cdi bean that talks to the openshift rest api in order to read/write the config maps. 

To do so we need to make some changes, like being able to hook back-end service to the data manager, and also integrate with the cache.

So far the pull request addresses two things:

- Align naming around kind, which sometimes is called model.
- Convert a plain singleton into CDI produces application scoped bean.
- Change the way we are handling caches, so that we can have a cache per kind, rather than a map per kind.
- Introduce the DataAccessObject interface
- Hook DataAccessObject with DataManager
- Implement Integrations DataAccessObject.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -65,6 +65,25 @@
     </dependency>
 
     <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+    </dependency>
+
+    <!-- Testing Dependencies -->
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>mockwebserver</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
--- a/api/src/main/java/com/redhat/ipaas/api/v1/model/Integration.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/model/Integration.java
@@ -28,6 +28,21 @@ public interface Integration extends WithId<Integration>, WithName, Serializable
 
     String KIND = "integration";
 
+    /**
+     *Required Labels
+     */
+    String LABEL_NAME = "ipaas.redhat.com/integration-name";
+
+    /**
+     * Optional Labels
+     */
+
+    //The integration id
+    String LABEL_ID = "ipaas.redhat.com/integration-id";
+
+    //The integration template id
+    String LABEL_TEMPLATE_ID = "ipaas.redhat.com/template-id";
+
     @Override
     default String kind() {
         return KIND;
--- a/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObject.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObject.java
@@ -1,0 +1,57 @@
+package com.redhat.ipaas.api.v1.rest;
+
+import com.redhat.ipaas.api.v1.model.ListResult;
+import com.redhat.ipaas.api.v1.model.WithId;
+
+public interface DataAccessObject<T extends WithId> {
+
+    /**
+     * @return The Type of entity it supports.
+     */
+    Class<T> getType();
+
+    /**
+     * Fetches an entity by id.
+     * @param id    The id.
+     * @return      The matching entity.
+     */
+    T fetch(String id);
+
+    /**
+     * Fetches a {@link ListResult} containing all entities.
+     * @return  The {@link ListResult}.
+     */
+    ListResult<T> fetchAll();
+
+    /**
+     * Creates a new entity.
+     * @param entity    The entity.
+     * @return          The created entity.
+     */
+    T create(T entity);
+
+
+    /**
+     * Updates the specified entity.
+     * @param entity    The entity.
+     * @return          The previous value or null.
+     */
+    T update(T entity);
+
+
+    /**
+     * Delete the specified entity.
+     * @param entity    The entity.
+     * @return          True on successful deletion, false otherwise.
+     */
+    boolean delete(WithId<T> entity);
+
+
+    /**
+     * Delete the entity with the specified id.
+     * @param id        The id of the entity.
+     * @return          True on successful deletion, false otherwise.
+     */
+    boolean delete(String id);
+
+}
--- a/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObject.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObject.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.redhat.ipaas.api.v1.rest;
 
 import com.redhat.ipaas.api.v1.model.ListResult;
--- a/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObjectRegistry.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObjectRegistry.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.redhat.ipaas.api.v1.rest;
 
 import com.redhat.ipaas.api.v1.model.WithId;
--- a/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObjectRegistry.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObjectRegistry.java
@@ -1,0 +1,45 @@
+package com.redhat.ipaas.api.v1.rest;
+
+import com.redhat.ipaas.api.v1.model.WithId;
+import com.redhat.ipaas.api.v1.rest.exception.IPaasServerException;
+
+import java.util.Map;
+
+public interface DataAccessObjectRegistry {
+
+    Map<Class, DataAccessObject> getDataAccessObjectMapping();
+
+    /**
+     * Finds the {@link DataAccessObject} for the specified type.
+     * @param type  The class of the specified type.
+     * @param <T>   The specified type.
+     * @return      The {@link DataAccessObject} if found, or null no matching {@link DataAccessObject} was found.
+     */
+    default <T extends WithId> DataAccessObject<T> getDataAccessObject(Class<T> type) {
+        return getDataAccessObjectMapping().get(type);
+    }
+
+
+    /**
+     * Finds the {@link DataAccessObject} for the specified type.
+     * @param type  The class of the specified type.
+     * @param <T>   The specified type.
+     * @return      The {@link DataAccessObject} or throws {@link IPaasServerException}.
+     */
+    default <T extends WithId> DataAccessObject<T> getDataAccessObjectRequired(Class<T> type) {
+        DataAccessObject dao = getDataAccessObjectMapping().get(type);
+        if (dao != null) {
+            return (DataAccessObject<T>) getDataAccessObjectMapping().get(type);
+        }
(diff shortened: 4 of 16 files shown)


KurtStam:
@iocanel any chance you can do a little writeup with a diagram? This so I can wrap my head around what it is you are trying to do :).

iocanel:
@KurtStam: I am trying to get started with implementing the backend for Integrations.

iocanel:
@KurtStam: added some more stuff in the description.

KurtStam:
@iocanel Thanks, that helps a little. If you make a diagram I would be able to give more constructive feedback though.

api/src/main/java/com/redhat/ipaas/api/v1/model/Integration.java
```diff
@@ -28,6 +28,13 @@
 
     String KIND = "integration";
 
+    String LABEL = "com.redhat.ipaas/integration";
```
jimmidyson: Also can you document what each label represents? What's the difference between `LABEL` & `LABEL_ID` for example?

api/pom.xml
```diff
@@ -65,6 +65,26 @@
...
 
     <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+      <version>${kubernetes.client.version}</version>
```
jimmidyson: Can we put these in `dependenciesManagement` in parent pom please so we don't need to spread version stuff through child poms?

api/pom.xml
```diff
@@ -65,6 +65,26 @@
...
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>mockwebserver</artifactId>
+      <version>${mockwebserver.version}</version>
```
jimmidyson: `dependenciesManagement`

api/src/main/java/com/redhat/ipaas/api/v1/model/ListResult.java
```diff
@@ -40,7 +40,7 @@
      */
     List<T> getItems();
 
-    class Builder<T> extends ImmutableListResult.Builder<T> {
+    public class Builder<T> extends ImmutableListResult.Builder<T> {
```
jimmidyson: Isn't `public` redundant here?

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationDAO.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private static final boolean containsIntegration(ConfigMap configMap) {
+        return configMap != null && configMap.getMetadata().getLabels().containsKey(Integration.LABEL);
+    }
+
+    private static final Integration toIntegration(ConfigMap configMap) {
+        if (configMap == null) {
+            throw new IllegalArgumentException("ConfigMap cannot be null.");
+        } else if (configMap.getMetadata() == null) {
```
iocanel: With your suggestion you have to read inside the block, to understand the flow.

No real difference though, I'll adopt your suggestion.

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClient;
+
+@ApplicationScoped
+public class KubernetesClientProducer {
+
+    @Produces
+    public KubernetesClient create() {
+        return new DefaultKubernetesClient();
```
iocanel: This needs to be addressed to the client side.

If we break the Config object, in two parts:
- properties that affect the connection _(e.g. url, keys, certificates)_
- properties that don't affect the connection _(token, credentials, timeout values)_

we could possibly allow partial client reconfiguration:

*DSL style*

     client.withRequstConfiguration(config).pods().list();

*Functional Style*

    PodList list =  client.withRequstConfiguration(config, new Function<KubernetesClient, PodList> {
            public PodList apply(KubernetesClient client) {
                  return client.pods().list();
            }
     });

*Lamda syle*

    PodList list =  client.withRequstConfiguration(config, c -> c.pods.list());


Eitherway, I think that this is outside the scope of this pull request.

pom.xml
```diff
@@ -112,6 +112,8 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
 
+    <kubernetes.client.version>1.4.30</kubernetes.client.version>
```
iocanel: And by the time I upgrade to 33 a new one will be around :-P

api/pom.xml
iocanel: Right!

api/pom.xml
iocanel: Right

iocanel:
Applied changes, re based and squashed.

iocanel:
@jimmidyson: If you can review this and merge, so that we can get the DataAccessObject in master, maybe I could hack a ConnectionsDAO, that would read integrations, parse the flow model and try to figure out connections n stuff.

jimmidyson:
@iocanel If you could get it to compile first then I might merge it :-b

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
jimmidyson: It is. I've raised #66 to track it.

iocanel:
@jimmidyson: build issues resolved!

jimmidyson:
Does the cache actually give us anything here or should we remove the cache & add it back in later once we've thought about caching properly?

iocanel:
Big discussion, too early in the process!

When accessing stale data, becomes an actual problem and when we have a couple of `DataAccesObject` implementations available _(with different requirements)_ we can revisit.

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationDAO.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Override
+    public Integration create(Integration entity) {
+        return toIntegration(kubernetesClient.configMaps().create(toConfigMap(entity)));
+    }
+
+    @Override
+    public Integration update(Integration entity) {
+        String id = entity.getId().orElseThrow(IllegalArgumentException::new);
```
jimmidyson: No description in exception - not good for caller.

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationDAO.java
```diff
@@ -0,0 +1,155 @@
...
+            return toIntegration(old);
+        } else {
+            return null;
+        }
+    }
+
+    @Override
+    public boolean delete(WithId<Integration> entity) {
+        try {
+            return kubernetesClient.configMaps().withName(entity.getId().orElseThrow(IllegalArgumentException::new)).delete();
```
jimmidyson: Same problem, no message in exception

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationDAO.java
```diff
@@ -0,0 +1,155 @@
...
+    public Integration create(Integration entity) {
+        return toIntegration(kubernetesClient.configMaps().create(toConfigMap(entity)));
+    }
+
+    @Override
+    public Integration update(Integration entity) {
+        String id = entity.getId().orElseThrow(IllegalArgumentException::new);
+        ConfigMap old = kubernetesClient.configMaps().withName(id).get();
+
+        kubernetesClient.configMaps().withName(id).replace(toConfigMap(entity));
```
jimmidyson: Won't this fail if previous get shows target didn't exist?

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObject.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    T update(T entity);
+
+
+    /**
+     * Delete the specified entity.
+     * @param entity    The entity.
+     * @return          True on successful deletion, false otherwise.
+     */
+    boolean delete(WithId<T> entity);
```
jimmidyson: Should we have a delete that just takes an ID?

jimmidyson:
Sorry for the bitty review, trying to squeeze it in when I get a moment over the weekend...

iocanel:
No worries! We'll do it at the pace we can :-)

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataAccessObjectRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param type  The class of the specified type.
+     * @param <T>   The specified type.
+     * @return      The {@link DataAccessObject} or throws {@link IPaasServerException}.
+     */
+    default <T extends WithId> DataAccessObject<T> getDataAccessObjectRequired(Class<T> type) {
+        DataAccessObject dao = getDataAccessObjectMapping().get(type);
+        if (dao != null) {
+            return (DataAccessObject<T>) getDataAccessObjectMapping().get(type);
+        }
+        throw new IPaasServerException("No data access object found for type: [" + type + "].");
```
jimmidyson: `IllegalArgumentException` perhaps?

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -57,68 +59,78 @@
...
     public void init() {
-        if (cache.isEmpty() && dataFileName != null) {
+        if (dataFileName != null) {
             ReadApiClientData reader = new ReadApiClientData();
             try {
                 List<ModelData> mdList = reader.readDataFromFile(dataFileName);
                 for (ModelData modelData : mdList) {
                     addToCache(modelData);
                 }
             } catch (Exception e) {
```
jimmidyson: Should we rethrow this error after logging so we fail fast if we don't read the data successfully?

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -182,74 +194,106 @@ public String generatePK(Map<String, WithId> entityMap) {
             default:
                 break;
         }
-        throw new ClassNotFoundException("No class found for model " + model);
+        throw IPaasServerException.launderThrowable(new ClassNotFoundException("No class found for model " + kind));
```
jimmidyson: From https://docs.oracle.com/javase/7/docs/api/java/lang/ClassNotFoundException.html, `ClassNotFoundException` probably isn't the right thing to throw here. How about `IllegalArgumentException` here too?

api/src/main/java/com/redhat/ipaas/api/v1/rest/Integrations.java
```diff
@@ -18,12 +18,18 @@
...
+import javax.inject.Named;
 import javax.ws.rs.Path;
 
 @Path("/integrations")
 @Api(value = "integrations")
 public class Integrations extends BaseHandler implements Lister<Integration>, Getter<Integration>, Creator<Integration>, Deleter<Integration>, Updater<Integration> {
 
+    @Inject
+    @Named(Integration.KIND)
+    private IntegrationDAO dao;
```
jimmidyson: I'm missing something - why can't I see where this is actually used?

iocanel: Right! This needs to go! The dao is only accessed via the datamanager.

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationDAO.java
```diff
@@ -0,0 +1,159 @@
...
+    public Integration fetch(String id) {
+        return toIntegration(kubernetesClient.configMaps().withName(id).get());
+    }
+
+    @Override
+    public ListResult<Integration> fetchAll() {
+        ConfigMapList list = kubernetesClient.configMaps().withLabel(Integration.LABEL_NAME).list();
+        List<ConfigMap> maps = list != null ? list.getItems() : Collections.emptyList();
+        List<Integration> integrations = maps.stream().map(c -> toIntegration(c)).collect(Collectors.toList());
+        return new ListResult.Builder<Integration>().items(integrations).build();
```
iocanel: No need to set total count here. This needs to be fixed elsewhere :-P

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -182,74 +194,106 @@ public String generatePK(Map<String, WithId> entityMap) {
...
-        if (entityMap == null || !entityMap.containsKey(id))
-            throw new EntityNotFoundException("Can not find " + model + " with id " + id);
-        //TODO interact with the back-end system
-        entityMap.remove(id);
+        if (!cache.containsKey(id))
+            throw new EntityNotFoundException("Can not find " + kind + " with id " + id);
+
+        WithId entity = cache.get(id);
+        if (entity != null && doWithDataAccessObject(kind, d -> d.delete(entity))) {
+            cache.remove(id);
```
iocanel: I like this idea, we should definitely try it out. Still, I see no harm in removing it here.

jimmidyson: Agreed, for now.

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -182,74 +194,106 @@ public String generatePK(Map<String, WithId> entityMap) {
...
+            return true;
```
iocanel: We have deleted the item succesfully from the backend, so we need to return true.

jimmidyson: So this again comes to cache staleness, what if `cache.remove` fails...? We'll need to revisit this in future, but for now let's get this merged.

jimmidyson:
LGTM [merge]

PR #67, opened 2017-01-24
Introduce Integration Patterns DAO

iocanel:
This pull reuqest adds a readonly DAO for integration patterns. 

Patterns are loaded from the funktion model. (stepkinds -> integration patterns).

Along with this the pull request also fixes 3 side issues:
- broken injection of cachecontainer
- broken injection of list of daos
- fetchAll not working correctly.

### DIFF
--- a/api/src/main/java/com/redhat/ipaas/api/v1/rest/CacheManagerProducer.java
+++ b/api/src/main/java/com/redhat/ipaas/api/v1/rest/CacheManagerProducer.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.redhat.ipaas.api.v1.rest;
 
 import org.infinispan.configuration.cache.Configuration;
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/redhat/ipaas/api/v1/rest/CacheManagerProducer.java
```diff
@@ -0,0 +1,48 @@
...
+        GlobalConfiguration g = new GlobalConfigurationBuilder()
+            .clusteredDefault()
+            .transport()
+            .clusterName("InfinispanCluster")
+            .build();
+
+        Configuration cfg = new ConfigurationBuilder()
+            .eviction()
+            .strategy(EvictionStrategy.LRU)
+            .maxEntries(150)
```
jimmidyson: Can we make this configurable?

api/src/main/java/com/redhat/ipaas/api/v1/rest/IntegrationPatternDAO.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public ListResult<IntegrationPattern> fetchAll() {
+        return new ListResult.Builder<IntegrationPattern>()
+            .items(PATTERNS.values())
+            .totalCount(PATTERNS.size())
+            .build();
+    }
+
+    @Override
+    public IntegrationPattern create(IntegrationPattern entity) {
```
jimmidyson: Shame we have to add this all in here, we can probably tidy up interfaces here somehow, but anyway at least we've done this in the entrypoint to the REST API so these should hopefully never be called.

api/src/main/java/com/redhat/ipaas/api/v1/rest/CacheManagerProducer.java
iocanel: Yes, we can use deltaspike's configuration value. But let's worry about it when we know what to do with the cache.

jimmidyson: OK

jimmidyson:
[merge]

PR #69, opened 2017-01-28
Various Docker updates

jimmidyson:
* Prefer ipv6 stack to prevent problems on ipv6 enabled servers
* Use JBoss base image to build on
* Use correct name for image
* Use latest fabric8 maven plugin
* Push image to docker hub from circleci

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -65,6 +65,15 @@
         <groupId>io.fabric8</groupId>
         <artifactId>fabric8-maven-plugin</artifactId>
         <version>${fabric8.maven.plugin.version}</version>
+        <configuration>
+          <generator>
+            <config>
+              <wildfly-swarm>
+                <name>rhipaas/ipaas-rest:%l</name>
+              </wildfly-swarm>
+            </config>
+          </generator>
+        </configuration>
       </plugin>
 
       <plugin>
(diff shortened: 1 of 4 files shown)


runtime/pom.xml
```diff
@@ -36,6 +36,9 @@
     <apidocs.output.dir>${project.build.outputDirectory}</apidocs.output.dir>
 
     <maven.build.timestamp.format>yyyy-MM-dd HH:mm</maven.build.timestamp.format>
+
+    <fabric8.generator.name>rhipaas/ipaas-rest:%l</fabric8.generator.name>
+    <fabric8.generator.from>fabric8/java-jboss-openjdk8-jdk:1.2.3</fabric8.generator.from>
```
rhuss: Actually I'd prefer to use XML configuration for this as its more typesafe. For changing parts like version one could still use properties.

E.g. 

```xml
<configuration>
  <generator>
    <config>
      <wildfly-swarm>
        <name>rhipaas/ipaas-rest:%l</name>
        <from>fabric8/java-jboss-openjdk8-jdk:${version.java-image}</from>
      </wildfly-swarm>
    </config>
  </generator>
</configuration>
```

Its admittedly more verbose and connects directly to `wildfly-swarm` connector whereas the property approach works for all java based generator, e.g. when switching to spring-boot, properties would work, too (for this two props).

Another point: I would use as version `1.2` so that we are free to increase the patchlevel of the image without the need to adapt the pom.xml. Again, somewhat discussable as it is a trade off between reproducibility and flexibility.

jimmidyson:
Thanks @rhuss. I'll make those suggested changes before we merge.

runtime/pom.xml
jimmidyson: Updated as recommended.

jimmidyson:
[merge]ing after following @rhuss' suggestions.

PR #71, opened 2017-01-30
First thoughts on a connection tester

rhuss:
This document describes first idea how a _connection tester_ could be implemented in the context of `iPaas`. Its all about providing feedback to the UI of whether a connector is functional with a given connector configuration.

Please feel free to comment on this document as review comment.


docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+When the user triggers this test function from within the UI, a backend call to `ipaas-rest` will be issued, which in turn will trigger a (still to be defined) method on the referenced connector with the concrete configuration given. This method, let's call it `ping()`, must be side effect free, which should be possible for most endpoints. The method takes the configuration as argument for the connector and returns a result objects which contains the result of the check (ok / nok) and in case of a failure the error message (question: Should we already think about I18N here ?)
+
+## Side effect free connection test
+
+Endpoints like database or REST endpoints can be checked side-effect free (e.g. for HTTP based services on could try a `HEAD` request to the root context). If this is not possible because e.g. some payload would be needed to be transferred in order to determine whether the connection works, then a connection test in the sense described here can't be performed. Instead some "sample flow run" needs to be performed which moves a test payload through the flow where the connector is at the the end (but this should be also represented differently on the UI).
```
davsclaus: I think this kind of logic is Camel component specific how they do a `ping` check. Not something that the ipaas project should deal with.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+Its important that the UI knows whether a connection is testable or not so that it can provide the appropriate UI state (e.g. an enable or disabled test button). Therefore each `Connector` has an extra `canPing()` method to allow to query for this information. `ipaas-rest` stores this information along with the `Connection` object which is handed out to the UI for showing the connection (along with the 'test' button). The method `canPing()` should receive the same `ConnectorConfiguration` as `ping()` itself so that it can dynamically determine whether a ping is possible. By default this argument to `canPing()` is probably ignored since a connector's singable-attribute is unique to the type of connector and independent of connection parameters.
```
davsclaus: Yes the Camel health-check needs a way to expose whether a component/connector supports `ping` or not. And maybe also whether it has `healthy` checks or not. I would however be a bit surprised if we have components that only have ping and not healthy checks and vice-versa. But lets find out on the way when we implement the code for this.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+In summary, `ping()` and `canPing()` are suggested as additions to the connector interface within the upstream Camel connector API definitions.
+
+## Connection Tester Runtime
```
davsclaus: Yeah all this pool of pods is similar idea to kubernetes fission: http://blog.kubernetes.io/2017/01/fission-serverless-functions-as-service-for-kubernetes.html - And I think @jstrachan also had some ideas about this for the fabric8 funktion or earlier when he was more involved with ipaas.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+It is essential, that connection testing happens as fast as possible as it is a synchronous operation. The initial trigger is a HTTP request from `ipaas-ui` to `ipaas-rest` which then performs the test action against the backend. A naive approach would be to spin up a Java pod with the referenced connector, call the `ping()` method, return the result to the client and (asynchronously) tears down the pod. However, the startup time are considered to be too expensive, to that it make sense to have a pool of pods running which can be directly contacted to perform the ping check. Various scenarios are conceivable:
+
+* **All-in-one** pods which can perform the ping check for all connectors. These would have access to all connectors supported and expose a service which has the connector type as parameter as well as the `ConnectorConfiguration` for calling out to the connector. A single K8s service is in front of these pods which is called by the `ipaas-rest`. This will work only if all connectors are known upfront or connectors could be loaded dynamically.
```
davsclaus: For the ipaas demo we could have this approach with a already warm pod with all the possible connectors.

davsclaus: Notice that you cannot likely have all 200+ Camel components in the same JVM without classpath classes and whatnot. So this can only take us so far. But for the first ipaas versions in 2017 it seems an okay approach.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+* **All-in-one** pods which can perform the ping check for all connectors. These would have access to all connectors supported and expose a service which has the connector type as parameter as well as the `ConnectorConfiguration` for calling out to the connector. A single K8s service is in front of these pods which is called by the `ipaas-rest`. This will work only if all connectors are known upfront or connectors could be loaded dynamically. 
+* **Dedicated** pods which holds only one connector each. Each of this pod has a different service as entry point (whose name is associated with the connector name). So there is one pod, one service and `ipaas-rest` will discover the proper service (by naming convention to the connector name) and call it with the `ConnectorConfiguration`. The advantage is, that if no such service could be found, the `ipaas-rest` could install the service on the fly. This works also for connectors which are not known upfront but might be defined later (e.g by expert developers).
```
davsclaus: We could create the pod when the user have selected the from connector in the ipaas US. Then its already booting up while the user is working in the UI. Then the pod is likely almost ready then the user may attempt to validate the connection. And as a user you are more inclined to accept a delay in such an operation as the user expects some backend / computer stuff to happen which can take longer time.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Open Questions
+
+* Is the set of connectors static or can these be defined by _expert developers_, too ?
+* If connectors can be defined dynamically, how ...
+    * ... are these build ?
+    * ... are they distributed ? (via Maven repo ?)
+    * ... how can they be loaded by the tester pods ?
+* How should I18N works for the error messages returned ? Is each connector responsible for I18N (and the locale should be part of the API) ? Or is I18N out of scope ?
+* Can connectors be tested with out a Camel route ? (would be the easiest)
```
davsclaus: Yes you can test Camel components without routes. But a route is often easier as its a natural way of using Camel.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+On a first glance, the fist **all-in-one** option is the easiest one, as it only needs a single service, where the pods behind will either already have all connectors preloaded or can load dynamically.
+
+## Open Questions
+
+* Is the set of connectors static or can these be defined by _expert developers_, too ?
+* If connectors can be defined dynamically, how ...
+    * ... are these build ?
+    * ... are they distributed ? (via Maven repo ?)
+    * ... how can they be loaded by the tester pods ?
+* How should I18N works for the error messages returned ? Is each connector responsible for I18N (and the locale should be part of the API) ? Or is I18N out of scope ?
```
davsclaus: I think i18n is out of scope currently. Lets wait for PM to put that on the table.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+On a first glance, the fist **all-in-one** option is the easiest one, as it only needs a single service, where the pods behind will either already have all connectors preloaded or can load dynamically.
+
+## Open Questions
+
+* Is the set of connectors static or can these be defined by _expert developers_, too ?
+* If connectors can be defined dynamically, how ...
```
davsclaus: Good questions. I think Jimmi, Keith et all needs to decide about this.

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+    * ... are these build ?
+    * ... are they distributed ? (via Maven repo ?)
+    * ... how can they be loaded by the tester pods ?
+* How should I18N works for the error messages returned ? Is each connector responsible for I18N (and the locale should be part of the API) ? Or is I18N out of scope ?
+* Can connectors be tested with out a Camel route ? (would be the easiest)
+
+## API suggestion
+
+```java
+interface Pingable {
```
davsclaus: At upstream Apache Camel its important that this is part of the health check api for components, and therefore the api will be component specific. The connectors will then simply inherit this functionality from the component they are based upon

docs/design/connection-tester.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Communication flow
+
+When the user triggers this test function from within the UI, a backend call to `ipaas-rest` will be issued, which in turn will trigger a (still to be defined) method on the referenced connector with the concrete configuration given. This method, let's call it `ping()`, must be side effect free, which should be possible for most endpoints. The method takes the configuration as argument for the connector and returns a result objects which contains the result of the check (ok / nok) and in case of a failure the error message (question: Should we already think about I18N here ?)
```
jimmidyson: We have no requirement for i18n right now, I'm assuming so far that everything will be in English (British English of course :-b).

Wonder what @kcbabo's thoughts on this are?

jimmidyson: `ping` to my mind relates to an ICMP request. Personally I'd be happy with a more verbose/expressive name. `validateConnection` might be better I guess?

docs/design/connection-tester.md
jimmidyson: Right - the ipaas should just be calling the relevant `ping` method & handle the response. It is up to the Camel connector to perform whatever it needs to do to validate connection,

docs/design/connection-tester.md
jimmidyson: I think having the separation of ping & healthy checks is useful. I see healthy checks as runtime checks, is stuff running fine? I see ping checks as preflight checks to make sure configuration is correct.

docs/design/connection-tester.md
jimmidyson: I've chatted to OpenShift team about something generic for this but there isn't anything we can use right now. Because we're only starting off without a curated set of pre-canned connectors, the all-in-one approach will work fine for now, but in time we need to figure this out.

docs/design/connection-tester.md
jimmidyson: Right. Any funky classloader stuff can wait.

docs/design/connection-tester.md
jimmidyson: Yes this sounds like a good approach for now, but I'd like to figure out how we can ensure we keep utilisation as low as possible. One problem we have is when to idle pods, how we track requests to pods and know if they are no longer being used - we can't rely on clean client closes. Also we have to look at how we can try to ensure pods are spun up as quickly as possible, remembering the eventually consistent, async nature of create pod requests. And then we have to handle pod crashes, etc... what's the UX in this case?

docs/design/connection-tester.md
jimmidyson: This is out of scope for now, curated connectors for preview.

docs/design/connection-tester.md
davsclaus: Yeah sure but we should add both concepts to Apache Camel. Doing a health check may involve some parts of what a ping would check, eg for a database to do a `SELECT * FROM DUAL`.

docs/design/connection-tester.md
davsclaus: We do have camel-grape that can load dynamic but its not so much used/tested so there can be problems with this approach. But a limited test is the IDEA plugin which uses it to load dynamic camel catalogs based on the version of Camel the user is using in the pom.xml file. Also the IoT guys are using it in the rhiot project.

docs/design/connection-tester.md
KurtStam: @rhuss I would want to add a version of test that is not 'side effect free'; it would actually invoke the 'twitter' endpoint as I think it's important that the enduser can test this for debugging purposes. He or she can then check that it actually worked. We could offer a 'undo' button if that's possible, or let the user manually undo their action if needed.

rhuss: @KurtStam I see the use case, but I would keep that separate for the moment, mostly also because you can't really define an API which is uniform over all connectors that takes a payload.

I think this use case is better covered when you test a whole integration. Testing the connector alone _functionally_ (i.e. doing what it is supposed) shouldn't be the task of the user but of us. 

So I think we should distinguish between:

* Testing whether a given configuration can connect with the backend. When testing a connection the user should be safe that there no changes will happen so that he can press the button as often as he wants.
* Testing an integration with a test message so that the user can manually verify the result. Sending a test message will always have a side effect, so the user knows this when pushing that button (which btw. should appear somewhere else in UI as it is a different use case),

rhuss: @jimmidyson I'm fine with `ping`,  `validate` (would skip the "Connection" as its clear from the context), `healthCheck`, `check`, `try`, ...  The important thing is that it should get a specific configuration as input, so its imo a bit different then a plain health-check which is without arguments so that it can be used by generic systems directly (like monitoring or runtime platform). A health check would be more appropriate for a `Connection` which already includes the config, but at this point we are dealing with generic `Connectors`.

@davsclaus can you give me some pointers to the Camel health check upstream efforts ?

docs/design/connection-tester.md
rhuss: Completely agreed, was probably stated a bit misleading. For me the important thing is that the API is uniform over all connectors, without knowledge how the actual check is done. Input is a concrete configuration (which is probable a generic typeless Map, with meaning specific for the connector ?), output is the ping result (ok/nok, error message if nok)

docs/design/connection-tester.md
rhuss: For me the big difference is, that a health-check can be called _without_ arguments (and works on a connection), a ping needs a _specific configuration_ as argument (and works on a connector), so that we can reuse the same `Connector.ping()` for many checks with different configurations (which also means that we can offer this as a stateless service).

docs/design/connection-tester.md
KurtStam: @rhuss I agree with the two types of tests, I just think the Citizen Developer will need this functionality now. We have no control over the 'twitter' endpoint; it can be intermittently down, it can have a bug, it can change over time, and potentially break the current integrations, etc, etc.

Also I think we do know the API contract at any point in the integration to allow for a data mapping step.

rhuss: @KurtStam actually, that's a very good point about changing endpoints. But I didn't think its the citizen developers task to detect this, because what can he do if he finds out that the connector is broken ? Its the operator's task (== we) that has to monitor the endpoints about changes. Also, the change most likely affects existing integrations not when you create a new one. But agreed, its an important thing to monitor.

The "connector test" happens without any integration or data mapping, but when you define a `Connection` (== `Connector` + user specific `ConnectorConfiguration`). The integration comes after this. The API contract between ipaas-rest and the `Connector` should be the same for every connector, each connector should know on its own how to verify a configuration. So its independent from any connector specific payload (which we cannot provide anyway when we define a connection, where the "Test connection" button is supposed to be placed).

docs/design/connection-tester.md
rhuss: @jimmidyson @davsclaus Just glanced over Fission. While not directly applicable (because of their language bias to Python and Node), their pod-pooling setup looks promising. I will play around with that tomorrow. Also there own (internal) routing service could be the extra indirection we might need to find pods holding specific connectors (via labeling).

jimmidyson: Yeah I discussed fission with @jstrachan briefly but we didn't get anywhere. Defniitely an interesting project to look at.

docs/design/connection-tester.md
davsclaus: @rhuss there is a JIRA at: https://issues.apache.org/jira/browse/CAMEL-10026. The work has not begun yet.

docs/design/connection-tester.md
davsclaus: Yes input as map should be possible so the parameters can be supplied using key/value pairs.

docs/design/connection-tester.md
KurtStam: @rhuss yeah so a connector will age, and when 2 or 3 years down the line the citizen developer starts using this connector it may no longer work as expected as let's say the twitter API slightly changed. So it's not only a thing to 'monitor', there should also be a test at creation time to test that the proposed solution is working; i.e. the return data format can have slightly changed.

Yes is happens before further integration but my point is that on a grand scheme we *know* what the message input and output is. Our plumbing can likely we abstract this message as some payload String; or "Connector + user specific ConnectorConfiguration", I'm sure we can send in an actual test message. I'm just saying this should be there in addition to what you are proposing. I'm happy to skip on it for this sprint, but I think we will need to have it before going out. We need to get this right, and we need to make it easy to build and validate each step as the citizen developer goes.

iocanel: I think that  `test` is more accurate than both `ping` or `validate`

rhuss: @KurtStam My hope is that when the backend changes we only need to adapt the connectors (and doing a rolling upgrade of all affected integration-apps with the new connector which has been rebuild), so that the use won't even recognise that there is a change in the backend API, if we keen the `Connector` API stable. Afterall that is one of the largest benefit of Camel that you don't have to care about backend specifics because you have a uniform (and stable) way to access components via Camel.

I'm still not convinced that sending a test message is needed for checking the 'connectiveness' of a Connector. However I'm absolutely convinced that we a test message flow during development of an integration as you described it.

My suggestion: Let's  implement the 'side-effect' free test first (as you said) and then decide about the test message flow. Is that ok for you, @KurtStam ?

rhuss: @KurtStam My hope is that when the backend changes we only need to adapt the connectors (and doing a rolling upgrade of all affected integration-apps with the new connector which has been rebuild), so that the use won't even recognise that there is a change in the backend API, if we keen the `Connector` API stable. Afterall that is one of the largest benefit of Camel that you don't have to care about backend specifics because you have a uniform (and stable) way to access components via Camel.

I'm still not convinced that sending a test message is needed for checking the 'connectiveness' of a Connector. However I'm absolutely convinced that we a test message flow during development of an integration as you described it.

My suggestion: Let's  implement the 'side-effect' free test first (as you said) and then decide about the test message flow. Is that ok for you, @KurtStam ?

docs/design/connection-tester.md
iocanel: Loading dynamically a function is one thing, loading camel components is an other.

docs/design/connection-tester.md
kcbabo: I agree on side-effect free test first.  I like the idea of the more evolved test Kurt is mentioning, but I think there's some bleed over into actions with that test.  IOW, I'm connecting and performing an action - that's the output you want to capture.  As of now, actions are configured independent of connections (in integrations), so we need more discussion around where this would go.

docs/design/connection-tester.md
kcbabo: Another thing to keep in mind with all-in-one or shared pods is multitenancy.  We are passing sensitive info (user/password/secret) in the connection configuration - we need to be careful about isolating that information and not allowing bleed through.

docs/design/connection-tester.md
kcbabo: +1

docs/design/connection-tester.md
kcbabo: For this particular aspect, yes.  For UI development in general, we should have an idea of how the console can be localizable, even if we are not localizing on our own.

rhuss: The problem is the I18N of error messages coming from the connection check which can be easily done in a generic fashion in the UI only but must be supported by the connector as each connector can create different error messages.

But that's just thought food for the future. Agreed that we shouldn't bother with this yet.

docs/design/connection-tester.md
rhuss: Yes, makes absolutely sense to have it on the `Component` interface, probably differentiating between producer and consumer (since you might only have read-only access to an endpoint).

davsclaus:
I logged ticket upstream: https://issues.apache.org/jira/browse/CAMEL-10795

rhuss:
Thanks @davsclaus ! Going to continue on the upstream for the API thing, will push an update here soon with the comments given.

docs/design/connection-tester.md
rhuss: @davsclaus for the tester, we probably don't need to reload routes but to instantiate `Connectors` to call their `test()` (or `ping()`) method. I wonder whether `camel-grape` is useful in this context, too (for dynamically loading connectors from somewhere) ?

rhuss: @kcbabo the 'connection test' services are _stateless_, to all context (including credentials) are provided as arguments to the `test()` method and are not stored anywhere. Therefore I don't think this poses a problem in a multi-tenant environment.

davsclaus: Yeah grape and camel-grape ought to be able to load any JAR dynamically if they are in maven central or some 3rd party maven repo. The latter requires to setup this. However we may need to improve camel-grape to support that, but we do that in the camel-catalog, and also in the IDEA plugin. However how that works running inside spring boot is not tested.

docs/design/connection-tester.md
rhuss: There needs to be a idle check for _tester pods_ so that they are scaled down (or removed) after a certain time of idleness. One could even play with some metrics (e.g. the time a test takes) measured to detect how _many_ pods should be kept around for serving testing request (horizontal auto scaling).

lburgazzoli:
I thought a little bit about how to introduce this "ping" concept on camel and I'm wondering if it could make sense for you guys to have an API like:

```java
enum ValidationScope {
    all,
    connectivity,
    ...
}

class ValidationResult {
    ValidationScope getValidationScope()
    String getErrorMessage()
    String getErrorCode()
    boolean isSuccess()
}

interface Component {
    ValidationResult validateOptions(ValidationScope scope, Map<String, String> options);
}
```

The goal is to be able to maybe add additional validation later on without changing the API.

rhuss:
@lburgazzoli what kind of additional scope would you think of ? Actually since the argument is already a typeless map, the implementation of such a map could do different checks based on the options. 

If you introduce such a scope, then you force all implementations to react on every possible value, possibly throwing an `UnsupportedOperationException` if not implemented. 

Actually I would keep the method as simple as possible, without a good reason for a premature abstraction. So without that scope. But feel free to convince me ;-) (maybe with a use case where different scopes are needed ?)

For the case of non-side effect testing where you send some payload to the component's endpoint one would need a different method anyway.

btw, I like @iocanel suggestion to use simply `test()` (and then with a `TestResult`).

iocanel:
Since we can't check the validity of the user configuration on in the UI _(unless we have means for letting the ui know the validation rules for each connector)_, it might make sense to have different `scopes` (e.g. configuration and connectivity validation). Not sure if the `scope` is the right word here ... 

Of course, I am not sure if we need to treat them separately or even if we need to expose them...

lburgazzoli:
The idea is that you may want to validate the options without connectivity, this may be used by tooling not leveraging the catalog and even camel itself to validate the options you set on the uri (sometime the stuffs you put in the uri options may be complex and not easy to validate through annotations). 

Then yeah, I do not know what is the best name: scope, context, etc

Btw, we did a little bit more "research" on possible api here: https://issues.apache.org/jira/browse/CAMEL-10795



https://issues.apache.org/jira/browse/CAMEL-10795

rhuss:
Let's continue the discussion on the API at https://issues.apache.org/jira/browse/CAMEL-10795

I will merge this document and update it accordingly.

PR #73, opened 2017-01-30
Migrate to Spring Boot

### LINKED ISSUE #72: Migrate to Spring Boot
Having tried valiantly to upgrade Wildfly Swarm to latest version and then add health checks to management port, I've just got bored... Wildfly Swarm may be technically better as a certain colleague of ours will argue, but it's just a pain to work with and feels really immature.

I've completed a migration to Spring Boot which was relatively painless. It actually builds and starts up faster, plus there are loads more resources to help us out when we get stuck.

I'm going to send in a PR in a minute, but this is just a heads-up for why. Please feel free to discuss, but I don't want to waste too much time before we migrate over so we can actually hit our deliverables for this sprint.

jimmidyson:
This migration does the minimum I could to make the switch, still using RESTEasy instead of Jersey/SpringMVC and still
using Infinispan. Both work great thanks to community starters from PayPal and Infinispan team.

Fixes #72


davsclaus:
Just a heads up that Spring Boot 1.5.1 was just released today. Apache Camel 2.19.0 has already been upgraded. But should work on the older 1.4.x release as well.

jimmidyson:
Thanks @davsclaus - already upgraded to 1.5.1 in this PR this morning.

KurtStam:
@jimmidyson So what are the reasons to go Spring Boot?

jimmidyson:
@KurtStam We've discussed this quite a few times, plus the issue I opened in #72 which eventually triggered me into action to do the migration. Biggest reason for migration is that firstly it just works, plus it has great documentation, many examples and a vibrant community. Personally I've found it hard to get started with Wildfly Swarm and have spent time fighting/figuring stuff out with it that would have been better spent actually coding more.

jimmidyson:
@davsclaus Downgraded to Spring Boot 1.4.4 when I tried adding Spring Cloud Sleuth for Zipkin tracing (breaking changes in Spring Boot 1.5.1 it seems that Cloud can't work with yet).

swagger-ui/src/main/resources/index.html
```diff
@@ -1,142 +0,0 @@
-<!--
```
jimmidyson: Why would we want the swagger UI in every deployment?

runtime/pom.xml
```diff
@@ -44,36 +44,43 @@
...
-        <groupId>org.wildfly.swarm</groupId>
-        <artifactId>wildfly-swarm-plugin</artifactId>
-        <version>${wildfly.swarm.version}</version>
-        <configuration>
-          <mainClass>com.redhat.ipaas.runtime.Main</mainClass>
-          <jvmArguments>
-            <jvmArgument>-server</jvmArgument>
-            <jvmArgument>-Xms256m</jvmArgument>
-            <jvmArgument>-Xmx512m</jvmArgument>
-          </jvmArguments>
```
jimmidyson: This doesn't really matter during development. Setting arbitrary limits here don't make much sense to me.

pom.xml
```diff
@@ -42,7 +42,7 @@
   <licenses>
     <license>
       <name>Apache License, Version 2.0</name>
-      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <url>http://www.apache.org/licenses/LICENSE-${javax.ws.rs-api.version}.txt</url>
```
jimmidyson: Apache 2.0 is correct - yeah my bad search/replace.

swagger-ui/src/main/resources/index.html
KurtStam: To facilitate integration and testing. It's great documentation format too. We wouldn't need to expose it in production if we don't want to.

runtime/pom.xml
KurtStam: It sure does matter even during development so it limits memory consumption. Plus it's a great placeholder to set these limits. It's much easier to debug a JVM to run out of memory then it is for OpenShift to blow itself up. Speaking from experience.

pom.xml
KurtStam: My point is though, maybe we want to be more restrictive with the license. This code is not 'framework' code. I think we don't really want to compete with another vendor who just takes the code and does not want contribute to the project.

swagger-ui/src/main/resources/index.html
jimmidyson: We expose swagger.json, isn't that enough?

runtime/pom.xml
jimmidyson: So where did this figure come from? Can you please share your calculations on how this amount of memory is reasonable for our REST service runtime? These are just arbitrary numbers so for now we should remove and refine it once we have some NFRs and some actual runtime numbers.

pom.xml
jimmidyson: Totally out of scope of this PR. Previous commits were under ASL2 and I'm happy to keep it as that anyway, but probably a broader discussion with PM. So we're not changing this in this PR - possibly in future following PM/BU input.

swagger-ui/src/main/resources/index.html
KurtStam: No, it's much nicer to have a dedicated swagger-ui; Apiman used embed the swagger-ui, and it's a great tool to display the swagger.json document as well as to interact with our REST API. It makes things easy for ourselves, QE and the UI team.

runtime/pom.xml
KurtStam: You're funny. We know the are going into battle but we don't know what type of bullets will be used, so we don't know what the thickness of the armor needs to be. So let's use no armor until we find that out. OK grrrreat.

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
```diff
@@ -15,17 +15,17 @@
...
-import javax.enterprise.inject.Produces;
-
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 
-@ApplicationScoped
+@Configuration
 public class KubernetesClientProducer {
```
iocanel: Maybe we should also rename from KubernetesClientProducer to KubernetesClientConfiguration.

or 

I would probably ditch that and add `spring-cloud-kubernetes` into the mix and get a fully configurable client for free.

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
```diff
@@ -15,17 +15,17 @@
...
 
-    @Produces
+    @Bean
     public KubernetesClient create() {
```
iocanel: same as above since we change style I would also renate from create() to kubernetesClient().

runtime/pom.xml
iocanel: I see no harm in having them set during development. We can tune as we go.

jimmidyson: No engineers are going to be hurt in the development of ipaas... well unless we get on and merge this quickly anyway...

But I can't be bothered to argue about this really - we've got functionality to add and this is an important change - so let's go with the arbitrary numbers.

swagger-ui/src/main/resources/index.html
iocanel: don't have a strong opinion on the subject ...

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
jimmidyson: Done the simple configuration switch for now. We can adopt spring-cloud-kubernetes if we need to later?

api/src/main/java/com/redhat/ipaas/api/v1/rest/KubernetesClientProducer.java
jimmidyson: Done.

swagger-ui/src/main/resources/index.html
jimmidyson: I see no reason to embed swagger ui in the REST server. It made loads of sense for APIMan which was all about managing APIs. This is not. We should document the swagger.json URL and recommend stuff like Swagger Chrome extension (https://chrome.google.com/webstore/detail/swagger-ui-console/ljlmonadebogfjabhkppkoohjkjclfai).

runtime/pom.xml
jimmidyson: Done.

jimmidyson:
Please final comments/review so we can merge this sucker today.

swagger-ui/src/main/resources/index.html
kahboom: What is the downside of keeping the swagger UI? The most important thing is the swagger.json as @jimmidyson said, but if the cost of keeping the swagger UI is exactly the same and maintenance is not an issue (just slows down deployment a bit), I don't see why keeping it is a big deal.

jimmidyson: Everything we add has a maintenance cost, regardless of size. It provides no useful function to the iPaaS itself and if it did I would want it deployed as a separate unit anyway, not embedded in here, following the single responsibility principle.

Regardless, I don't see this question blocking this reasonably large change migrating to Spring Boot. If people really feel strongly about it, then open an issue to discuss, but please don't block this PR due to swagger UI integration when there are a myriad of ways to use swagger ui, the simplest being the chrome extension I mentioned above.

rhuss: Isn't the swagger UI then also exposed to the end user ? If so, simply by its existence it already increase the attack surface considerably because since Swagger is a widely used framework this could make iPaaS vulnerable for automated attacks in case of an security breach.

So, imo because of maintenance and security we should keep it out in the _final version_. If we remove it now, we could be sure. Otherwise there is a chance we can forget to remove it.

jimmidyson: Yes it would be exposes to end user. As much as I'd love to use "it's a security issue" as an excuse not to add it, by its nature as a purely HTML/JS frontend that's not an argument that holds.

kahboom: @jimmidyson Gotcha. @KurtStam I think it could be worth it to keep the Swagger UI, but Jimmi is right in that it should be a separate GH issue and certainly shouldn't block this PR. The Chrome extension can be a temporary thing and we can look into providing our own later on since it's a lower priority for the MVP. I say get your merge on @jimmidyson .

rhuss: @jimmidyson there could be XSS issues if someone can control the content of a response (i.e. by a malicious API object's content), which then is displayed unescaped in Swagger UI (something  like in https://community.rapid7.com/community/infosec/blog/2016/09/02/r7-2016-19-persistent-xss-via-unescaped-parameters-in-swagger-ui). 

Not that I say that someone can easily control the content we send over our API, but Swagger UI could potentially used for XSS as long as it can show user generated content, which can lead to actions executed in the context of a connected user.

All I wanted to say, that it increases the attack surface, which could be easily avoided by dropping it when not really needed.

jimmidyson: Well that is an interesting one. So yeah - security says no ;)

kahboom: rofl

KurtStam:
[merge] please

runtime/pom.xml
KurtStam: Thanks!

PR #83, opened 2017-02-06
Add swagger integration test

jimmidyson:
Fixes up static swagger generation and adds a basic integration test to ensure it doesn't revert again.


pom.xml
```diff
@@ -311,7 +312,7 @@
           <version>${fabric8-maven-plugin.version}</version>
         </plugin>
         <plugin>
-          <groupId>com.github.kongchen</groupId>
+          <groupId>dk.nykredit.swagger</groupId>
```
rhuss: Just out of curiosity: Is this a different plugin or has the plugin been moved ? And it a change, what is the reason ?

runtime/src/test/java/com/redhat/ipaas/runtime/BaseITCase.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private TestRestTemplate restTemplate;
+
+    @Autowired
+    public void setRestTemplate(TestRestTemplate testRestTemplate) {
+        testRestTemplate.getRestTemplate().getMessageConverters().add(new YamlJackson2HttpMessageConverter());
+        this.restTemplate = testRestTemplate;
+    }
+
+    public TestRestTemplate restTemplate() {
```
rhuss: Still not keen on `get` methods ? ;-)

runtime/src/test/java/com/redhat/ipaas/runtime/BaseITCase.java
```diff
@@ -0,0 +1,58 @@
...
+    public static final SpringClassRule SPRING_CLASS_RULE = new SpringClassRule();
+
+    @Rule
+    public final SpringMethodRule springMethodRule = new SpringMethodRule();
+
+    private TestRestTemplate restTemplate;
+
+    @Autowired
+    public void setRestTemplate(TestRestTemplate testRestTemplate) {
+        testRestTemplate.getRestTemplate().getMessageConverters().add(new YamlJackson2HttpMessageConverter());
```
rhuss: Why do we need a YAML converter ?

pom.xml
jimmidyson: Previous plugin had problems with generating swagger document. New plugin actually works by simply delegating to Swagger's own libraries.

runtime/src/test/java/com/redhat/ipaas/runtime/BaseITCase.java
jimmidyson: Oops thought I'd removed that - will do so now.

runtime/src/test/java/com/redhat/ipaas/runtime/BaseITCase.java
jimmidyson: :) Fight!

runtime/src/test/java/com/redhat/ipaas/runtime/BaseITCase.java
jimmidyson: Ah I remember, this is only used for swagger.yaml (YAML twin of swagger.json). Only needed for that test.

jimmidyson:
[merge]

jimmidyson:
Thanks @rhuss!

Issue #82, opened 2017-02-06
Use a single style of maven properties


iocanel:
So is it....

version.xxx or xxx.version ?
xxx-yyy-version or xxx.yyy.version?
xxx-yyy.version or xxx.yyy.version?

and of course

xxx-maven-plugin.version or maven-xxx-plugin.version?

Personally, I don't care what it is as long as its a single style.

jimmidyson:
+1. Personally xxx-yyy.version and xxx-yyy-maven-plugin.version

iocanel:
xxx-yyy.version and xxx-yyy-maven-plugin.version it will be then

iocanel:
In other words <artfiactId>.version

jimmidyson:
Fixed by #84

PR #84, opened 2017-02-06
#82: Use <artifactId>.version properties for all dependencies and plu…

iocanel:
…gins. Sort alphabetically property names.


jimmidyson:
ok to test

PR #88, opened 2017-02-07
Issue74

KurtStam:
This PR set the 'real' data for the twitter component and connection. I've taken a first stab at the Salesforce connector meta data but work on that connector has not yet started, so things will get updated later there.

Note that for now the component.properties and the connection.configuredProperties are identical, since we don't look at the expert developer workflow just yet.


rhuss:
> @KurtStam, thanks! @KurtStam, please review this.

Hey @jimmidyson , your @jimmibot has been bribed ;-)

runtime/src/test/java/com/redhat/ipaas/runtime/ComponentsITCase.java
```diff
@@ -26,15 +26,15 @@
...
-        assertThat(results.getItems()).hasSize(20);
-        assertThat(results.getTotalCount()).isEqualTo(50);
+        assertThat(results.getItems()).hasSize(2);
+        assertThat(results.getTotalCount()).isEqualTo(2);
     }
 
     @Test
     public void componentsGetTest() {
-        Component result = restTemplate().getForObject("/api/v1/components/1", Component.class);
+        Component result = restTemplate().getForObject("/api/v1/components/org.foo:twitter-mention-connector:1.0", Component.class);
```
jimmidyson: I know it's a breaking change to the API, but shouldn't this be `connectors` rather than `components` now? Same change for the kind in the deployment.json too I would recommend.

KurtStam: 1. I'm all for keeping things consistent, though I had not heard we renamed Component to Connector, last I heard we were naming the 'Apache Connector' a 'Component' in the UI. 
2. When will be adding in 'actions'?
3. BTW, I don't think 'Components' are a part of the demo directly as we are focussing on Connections and Integrations for the demo? So maybe we should wait with renaming till after the demo?

jimmidyson: AFAIK there is no concept of Component in the UX. Can you share screens where that applies? As I sent in my email, we have the overloaded terms of Camel Connector and iPaaS Connector which have subtly different meanings sadly, but Component isn't in any design that I'm aware of.

KurtStam: Component = Connector + config (of all the possible options + their description). I think this is all Camel inspired terminology and it was adopted by the UX team who made the model. We simply adopted that model.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
```diff
@@ -21,30 +21,30 @@
 import java.io.Serializable;
 
 /**
- * ComponentGroups are labels in Camel.
+ * ConnectorGroups are labels in Camel.
```
rhuss: If they are labels in camels, why not calling it labels here, too ?

IMO a "group" is something that contains something, so an explicit way to group stuff. A "label" is something attached to an object, and the group of obects attached with the same label build an implicit grouping. 

You change a group by adding / removing references to the members (where maybe the member has even a reference to its parent groups), labels are much more decoupled.

I don't know the semantics which are supposed to be used here, but a "Group" and "Label" are quite different concepts which cannot be easily mapped frictionless in the backend.

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
```diff
@@ -146,10 +146,10 @@ public String generatePK(Cache<String, WithId> entityMap) {
 
     public Class<? extends WithId> getClass(String kind) {
         switch (kind.toLowerCase()) {
-            case "component":
-                return Component.class;
-            case "componentgroup":
-                return ComponentGroup.class;
+            case "connector":
```
rhuss: If `KIND` is already defined a `Connector.KIND`, it should be used here, too, as constant.

appveyor.yml
```diff
@@ -1,3 +1,19 @@
+#
+# Copyright (C) 2016 Red Hat, Inc.
```
rhuss: minor nit pit, but we should update our copyright profiles to the current year.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
jimmidyson: Agreed. These are not groups at all, but arbitrary strings AFAIK.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
```diff
@@ -21,30 +21,30 @@
...
  *
  * https://github.com/apache/camel/blob/master/platforms/catalog/src/main/java/org/apache/camel/catalog/CamelCatalog.java#L233
```
jimmidyson: This link is no longer valid. Please link to commit rather than master if we think this is useful (it probably is).

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
```diff
@@ -21,30 +21,30 @@
...
  *
  * @author kstam
  *
  */
 @Value.Immutable
-@JsonDeserialize(builder = ComponentGroup.Builder.class)
-public interface ComponentGroup extends WithId<ComponentGroup>, WithName, Serializable {
+@JsonDeserialize(builder = ConnectorGroup.Builder.class)
+public interface ConnectorGroup extends WithId<ConnectorGroup>, WithName, Serializable {
```
jimmidyson: Is this the way we map Camel connectors to iPaaS connectors?

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
jimmidyson: +1

appveyor.yml
jimmidyson: Should we?

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
KurtStam: The IPaaS UI defined Labels as an enduser defined concept. Also they defined a group so a ll the Twitter* Connectors can be mapped to one IPaaS Twitter Connection. We can push back on how the UX team thinks about it but they don't like changing things.. So that said, all the Twitter Connectors in Camel will be labeled with a twitter label, so we can use that to group them in IPaaS. Does that satisfy your concerns, or do you think we should push back to Camel or the UX team?

api/src/main/java/com/redhat/ipaas/api/v1/rest/DataManager.java
KurtStam: Yes that totally makes sense to me. Will update it.

appveyor.yml
KurtStam: I think you usually use the year of inception? Or else we need to update all the headers every year? I don't feel strongly either way though. Just like to keep the noise low.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
KurtStam: yes it is.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
KurtStam: ok - I've removed it.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
jimmidyson: Good point that we need to be able to have user defined labels but don't think ConnectorGroup is the right name.

appveyor.yml
jimmidyson: Agree with you @KurtStam, no reason to update copyright I don't think. 25 years or whatever it is should be enough.

KurtStam:
@gashcrumb @kahboom can you review the naming change from component to connector? This affects the REST API, so it will break things on your end when it gets merged in.

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
KurtStam: @jimmidyson  @rhuss The userDefined Labels are called "Tags", ConnectorGroups are driven by Camel Labels to Group all IPaasConnectors for let's say "Twitter". Let's not get hung up on this until we fully flesh out how Actions are going to work.

gashcrumb:
@KurtStam I think we'll actually be just fine on the UI side, at least as far as I can tell.

jimmidyson:
@gashcrumb Even with the change of URL for the REST call???

gashcrumb:
Well, I did qualify that statement :-)

appveyor.yml
rhuss: agreed, just wanted to pass the minimal 3 comment requirement for a review ;-P

rhuss: Since I suppose everyone uses some sort of automatism from its IDE to create such headers automatically when creating new files, i meant it as a heads up to update this personal IDE config.

nothing more :)

api/src/main/java/com/redhat/ipaas/api/v1/model/ConnectorGroup.java
rhuss: I probably have to dive deeper into the code, but is it correct when I assume that when the UI requests the `Connectors` for a  `ConnectorGroup` then ipaas-rest does a live-query on the `Connectors` which carry a certain label ? So a `ConnectorGroup` persistent attributes are the name of the group + the label for selecting the Camel connectors ?

Or in other words: A `ConnectorGroup` does *not* have a direct (db) reference to the `Connector` objects it contains in the DB (which needs to be maintained) ?

If so, I'm somewhat fine with calling it `ConnectorGroup` but handling it like a label.

jimmidyson:
I'll fix up conflicts and merge in ASAP.

jimmidyson:
Well I messed it up... this needs a rebase @KurtStam and remove my last commit :) As a maintainer you can't do a rebase and force push without the PR closing as I've learned from past experience...

jimmidyson:
Thanks @KurtStam! [merge]

PR #95, opened 2017-02-16
Disable CSRF as this is a bearer token only authenticated REST API

jimmidyson:


jimmidyson:
Self-[merge]ing as noone around to review.

PR #96, opened 2017-02-16
Use comma-separated allowed origins to appease Spring Boot yaml parser

jimmidyson:


jimmidyson:
Self-[merge]ing as noone around to review.

rhuss:
Sorry, my fault. But looks, although I wonder why Spring is so picky with one element lists in yaml.

PR #98, opened 2017-02-16
Serialize non-absent to work better with Java8 Optionals

jimmidyson:
Absent deals with empty Optionals as well as nulls


KurtStam:
@jimmidyson Not sure I understand what the problem is you are trying to solve. Happy to go along with it, I just don't really understand the change.

jimmidyson:
The existing serialization config is `JsonInclude.Include.NON_NULL`, which means don't include any fields  that are null in serialized JSON. This works great, but when you're using `Optional` typed fields and the field is non-null with an empty `Optional` value, the field is serialized as `null`. This change means that any absent valued fields (i.e. `null` or empty `Optional`) will not be included in JSON, which is consistent behaviour.

jimmidyson:
[merge]

PR #99, opened 2017-02-16
Saving to cache only

KurtStam:
@gashcrumb @jimmidyson if saving the integration keeps having issues like https://gist.github.com/gashcrumb/263c25b34e4d0bbf75139011622a3b9c
then we can disable saving to a configmap with this PR.


jimmidyson:
Let's [merge] this & think about how we persist stuff in next sprint

PR #104, opened 2017-02-17
Add support for using @AllowedOrigins("*") to do per class/method CORS configuration and use it so that http://petstore.swagger.io/?url=http://localhost:8080/api/v1/swagger.json loads ok.

chirino:


jimmidyson:
Merging on green. Thanks @chirino!

jimmidyson:
[merge]

PR #105, opened 2017-02-20
Adding a way to generate Server Side Events from the API module.

chirino:


chirino:
Will close out as this PR has gone stale and needs a rebase.

PR #107, opened 2017-02-20
Add an ipaas-ui client id to the keycloak realm

chirino:
Add an ipaas-ui client id to the keycloak realm so that the ./start-with-keycloak.sh script in this project can be used with the default ‘yarn start’ command in the ipaas-client and have the oauth just work. (cherry picked from commit a757b75234b3c50ff28e911dd997e652261c05e9)

### DIFF
--- a/runtime/src/test/resources/ipaas-keycloak-realm.json
+++ b/runtime/src/test/resources/ipaas-keycloak-realm.json
@@ -58,6 +58,37 @@
       "enabled": true,
       "publicClient": false,
       "bearerOnly": true
+    },
+    {
+      "clientId": "ipaas-ui",
+      "surrogateAuthRequired": false,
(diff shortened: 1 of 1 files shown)


KurtStam:
We now have 'ipaas-test' and 'ipaas-ui' clients. Is there any reason we shouldn't call 'ipaas-test' to 'ipaas-rest'? Or should both have some kind of 'test' in the name?

chirino:
@KurtStam I'm not really picky about what we call them.  ipaas-test is for the tests in the ipaas-rest, so perhaps it should be ipaas-rest-test.  ipaas-ui seems ok as is since it's used by the ipaas-ui project.

chirino:
retest this please

jimmidyson:
[merge]

PR #108, opened 2017-02-21
Consistent naming of integration ConfigMap labels

rhuss:
and inroduced enums to make it slightly more typesafe (and baked in the
comment 'mandatory' into the enum, not sure whether its needed).

Not even sure whether this is needed as the backend probably moves to a DB anyway.

Feel free to reject this PR :)


KurtStam:
Yeah not sure yet if we're going to keep this code so let's wait and see.

chirino:
is this PR still valid?

chirino:
@KurtStam any updates?

KurtStam:
I think we can close?

KurtStam:
Patch is out of date

PR #109, opened 2017-02-21
Reorganized internal package structure + some renaming

rhuss:
* Aligned Maven module and top-level package names
* Separated controller + DAOs
* Changed naming of controllers (Connections --> ConnectionHandler)
* Introduced domain specific packages
* Changed 'kind()' to 'getKind()' for consistencies sake (compared to other `With..` interfaces).

Comment to reviewers: A lot of stuff has been moved around, but not functionally changed. So its best to not look at the diffs, but review by browsing top-level through the directory structure.

It would be super awesome if we good do a review *fast* because otherwise integration with other PRs will get super hard then.


rhuss:
@jimmidyson  and all reviewers: Best to start here: https://github.com/rhuss/ipaas-rest/tree/reorg/rest/src/main/java/com/redhat/ipaas/rest

jimmidyson:
While I'm not 100% keen on the pretty deep package structure, it's definitely better than it was :) Let's [merge] for now and see how it feels.

PR #110, opened 2017-02-21
The dynamic application of the CorsFeature did not work on the OPTION http method for some reason. Back it out for now until we figure out why.

chirino:
The dynamic application of the CorsFeature did not work on the OPTION http method for some reason.  Back it out for now until we figure out why.
This reverts commit a26b9595e9010adfb1d4a18724a87fbe9717e820.


chirino:
Figured out how to fix it.. will submit a new PR for the fix instead of backing this out.

PR #113, opened 2017-02-21
Switch to Spring CorsFilter

jimmidyson:
@chirino @gashcrumb This works for me and seems a bit simpler.

PR #114, opened 2017-02-21
Honor the application.properties CORS config for OPTIONS requests

chirino:
Ok implement better default option handling for the CORS case.  Previous hack would report a 200 status even for nonexistent paths.  This fix is much better and honors the applicaiton.properties cors config for OPTIONS requests.


chirino:
Closing, I like the simplicity of PR #113 better

chirino:
Actually don't think #113 fully works.  Reopening this one for now.

jimmidyson:
#113 does work ;)

PR #118, opened 2017-02-23
Add a CacheForFilter and @CacheFor annotation to do annotation based http response cache configuration, it defaults to disable caching when not configured.

chirino:


jimmidyson:
Do we want browsers/clients to cache responses from our REST API? Not sure we do.

chirino:
@jimmidyson exactly.. we generally don't want them to cache the responses.  This filter disables caching by default unless you explicitly annotate the class/method with the '@CacheFor' annotation.  

Without this change we never set the Cache-Control header which let browsers like IE cache responses which caused problems.

KurtStam:
@chirino the things we do for IE.. @dsimansk do we have a list of browsers we have to support and if so what is it? BTW are we working towards running the automated tests on each of those?

jimmidyson:
@chirino Really?!?! That's awful!

chirino:
retest this please

jimmidyson:
[merge]

PR #119, opened 2017-02-23
Add initial database schema diagram and sql file

jimmidyson:
Might be easier to discuss on here, plus you can see the initial SQL file to create the database schema.

We'll need to look at migrations going forward, but the sooner we do that the better to ensure we have a good story for
real upgrades, when we get to that point.

/cc @rhuss @iocanel @KurtStam @kahboom @gashcrumb


rhuss:
@jimmidyson with which tool did you create the diagram  ?

docs/database/schema-alternative.sql
```diff
@@ -0,0 +1,101 @@
+--
```
rhuss: @jimmidyson this is the sql corresponding to https://github.com/redhat-ipaas/ipaas-rest/issues/91#issuecomment-282102839 Could you do me a favor and generate the ERP diagram for this, too ?
Thanks !

Not all indexes have been added though (and don't know whether its fully correct, haven't deployed it yet).

kahboom: @jimmidyson Once you're done with it can I make it pretty like the second diagram I made for James S? I'll color-coordinate it for you and everything. :D It's probably going to go on the iPaaS website once I'm done "revamping" it this week. lol

jimmidyson:
@rhuss Created in IDEA - just open the UML file.

docs/database/schema-alternative.sql
```diff
@@ -0,0 +1,101 @@
...
+
+CREATE EXTENSION hstore;
+
+-- ==== Instances ====================================================================
+
+-- An action instance is the single entry point in this sub domain. It is referenced
+-- by an integration and has all configuration needed for creating an instance of
+-- a fully configured Camel connector
+CREATE TABLE action_instance
```
jimmidyson: You're proposing we store the action configuration (effectively Camel connector configuration) in the database. Is this duplicated in the integration project YAML? Which is authoritative?

docs/database/schema-alternative.sql
jimmidyson: @rhuss Thank you for this alternative. I played with the separate connector_property and action_property at first and then changed it to the common table, with relations to action and connector. Looking at both options again, I think I prefer the separate tables as you have done, albeit that means we have two tables with identical column definitions, which is normally one thing I hate...

My only question is about the action_instance table and how this relates to an Camel connector configuration for an integration, correlating the two together somehow could be trickier if we have multiple sources for the same data - how do we handle them getting out of sync? Which one is authoritative?

KurtStam:
I've added https://github.com/redhat-ipaas/ipaas-rest/pull/120 UML fragments as a competing solution as I think that maps much better to the concepts we have been discussing.

docs/database/schema-alternative.sql
rhuss: This table is for me the combination of `configured_connector` configuration which is specific for a connector (mostly or exclusively defined by the expert dev) and the action configuration (defined by the citizen user) and the natural hand-over point to integrations. It contains all _configuration_ relevant information.

The alternative to keep the connector specific configuration in the DB and the action specific config in the integration YAML's is imo not favourable as you have to query now two different stores for getting a picture for the full configuration.

Also, I think that the configuration values in the YAMLs (not the structure) should come from the DB (as it is supposed to be changeable by a cit user, too, who has no idea of 'git'). So w.r.t parameter configuration, the DB is for me the authoritative source.

jimmidyson: While I agree that this approach seems to make more sense and I also prefer to have a single source of truth (the DB in this case), I still have a problem with how this can work once we have expert developers hacking on git repos. Do we defer that for now and figure it out when it becomes something we need to worry about? I'm happy to do that so we can get this moving.

docs/database/schema-alternative.sql
rhuss: I believe that `action_property` and `connector_property` might look incidentally the same but they are different _things_. E.g if one would like to add a flag whether a property is hidden or not, this could go only to `connector_property` (as I suppose that the action properties all have business meaning, and should be changeable by the cit user ?). Even if I wrong with my assumption one has still this extra degree of freedom to evolve these parts differently. 

Another, more technical argument is, that SQL debugging via relationship tables is always tedious, requiring two joins. Also you would have to add extra constraints to ensure that one config property only belongs to an action or a connector. And finally, looking in to this combine configuration_property table only, you can not decide whether a property is connector or action related. You would need to examine both relationship tables. And finally finally diamond structures are always evil ;-)

rhuss:
@jimmidyson hmm, whatever I try, I can't open the UML diagram in my IDEA (2016, 2017 EAP, ..., dark-backgorund, white-background). I always get an empty pane. Anything I would have to install in addition ?

And how could you create this out of an SQL schema ? Thought that IDEA only support UML class diagrams on Java code ...

docs/database/schema-alternative.sql
rhuss: I also opt for going on and keep the configuration in the DB for now.

One solution could be, that a expert dev only edits `template yamls` and the final yamls are generated artifacts which e.g. could be automatically regenerated . In this templates there are only the _reference to the configuration parameters_ stored (e.g the value could be ignored or some variable ref could be used). During the render process the DB is contacted and the values are filled in. This would also clearly separate what can be done by an expert dev (change the template) and by the cit user, too (define values to be filled in).

jimmidyson:
@rhuss I'm honestly not 100% sure how I did it, but I was definitely on latest 2017 EAP and used the Database view. I started up a PostgreSQL docker image like:

```bash
$ docker volume create pgdata
$ docker run --rm -e POSTGRES_PASSWORD=password -v `pwd`/docs/database/schema.sql:/docker-entrypoint-initdb.d/schema.sql:Z -v pgdata:/var/lib/postgresql/data -p5432:5432 postgres:9.6.2
```

Which starts up the server with the SQL file applied. Then used the Database view in intellij to view the diagram and export it... Seemed to work OK. I haven't tried opening the UML file actually, it was opened and saved, never closed and reopened... let me try now.

jimmidyson:
Yeah it just opens for me...

rhuss:
@jimmidyson thanks ! I haven't created a DB yet, will try this now, too.

docs/database/schema-alternative.sql
jimmidyson: You've convinced me ;)

jimmidyson:
After discussing with @rhuss, I've switched the schema to use his updates and updated diagrams accordingly.

Comments please. We need to get this DB schema sorted today so we can move forward.

rhuss:
lgtm :). Would we then introduce also same named objects with the corresponding relations to reflect this in our domain model ?

kahboom:
> lgtm :). Would we then introduce also same named objects with the corresponding relations to reflect this in our domain model ?

I think this would be very helpful when working on designs with the UXD team, and just for anyone that is using the project without being familiar with Camel (or who don't care to know what happens on the backend). I had previously brought this up to @jimmidyson, because it does sound like we need two separate diagrams to reference. @rhuss or do you think this can/should be in a single diagram?

jimmidyson:
Added property index columns for display and set more columns to not null.

jimmidyson:
Added integration tables.

docs/database/schema.sql
```diff
@@ -103,3 +103,28 @@ CREATE TABLE action_instance
...
+CREATE UNIQUE INDEX integration_name_uindex ON integration (name);
+
+-- An integration step is the definition of a single step in an integration. It contains configuration
+-- for each and every step in the integration flow.
+-- This can reference an action instance or be a supported integration pattern (e.g. log, wiretap, choice, etc).
+CREATE TABLE integration_step
+(
+    id SERIAL PRIMARY KEY NOT NULL,
+    integration_id INTEGER NOT NULL,
+    action_instance_id INTEGER,
```
rhuss: So I guess for intermediate steps this can be `null` ? Can an intermediate step refer an action  ? Probably not, so maybe we can model this in Java as `ActionStep extends Step` and leave the action_instance_id null for the base class.

gashcrumb:
So each step in an integration refers to an action, does then extend an action to mean a connector or integration pattern etc?

docs/database/schema.sql
```diff
@@ -103,3 +103,28 @@ CREATE TABLE action_instance
...
+    step_index SMALLINT NOT NULL,
+    step_type VARCHAR(32) NOT NULL,
```
rhuss: What kind of steps do we have ? In a hibernate world this would be the discriminator column of an inheritance hierachy, right ?

docs/database/schema.sql
```diff
@@ -103,3 +103,28 @@ CREATE TABLE action_instance
...
+
+-- An integration step is the definition of a single step in an integration. It contains configuration
+-- for each and every step in the integration flow.
+-- This can reference an action instance or be a supported integration pattern (e.g. log, wiretap, choice, etc).
+CREATE TABLE integration_step
+(
+    id SERIAL PRIMARY KEY NOT NULL,
+    integration_id INTEGER NOT NULL,
+    action_instance_id INTEGER,
+    step_index SMALLINT NOT NULL,
```
rhuss: That the order of step in the list of step, right ? would it make sense to explicitly chain steps (e.g. `previous_step` column). Just asking, no feeling whether index based or chained steps are easier to handle.

jimmidyson:
Each step references an optional (nullable action_instance_id) action. So it's either an action step or an integration pattern step.

docs/database/schema.sql
jimmidyson: Right.

docs/database/schema.sql
jimmidyson: I think it's easier to use index as then you can use SQL to order the results returned from the DB rather than having to recreate order in Java.

docs/database/schema.sql
jimmidyson: Yeah that's right, this is a discriminator, but in Java world it doesn't really have any need for inheritance hierarchy (I don't think) as they're simple steps. There could be an argument in future for step validation I guess where that kind of discriminator could be useful. This is more for the UI to identify what to display and also when we're creating the integration how to marshal this to a flow definition.

gashcrumb:
Sounds good.  So then the `step_type` field would be used to figure out what table to reference, i.e. whatever table holds integration patterns vs actions?

How would a child integration fit in?  Should there be a separate step type for that which would contain an id pointing to the child integration?

jimmidyson:
A child integration could be a new FK column integration_id to integration(id) in the future perhaps?

jimmidyson:
Oh and right now we don't have a table for integration patterns as for now I think we can get away with just hard coding functionality in UI for supported steps. This could change in future but I think this is the fastest way for us to get going and then think about integration patterns in DB later, if we need to.

Happy to think about that now if people think we should?

gashcrumb:
I think that'll be fine for now, provided we can build up the step array/tree properly in the API implementation.

wonder, maybe if step_type is 'integration pattern' or whatever you might need a reference ID or something so that the implementation can figure out which integration pattern that step refers to?  In fact, maybe there should just be a generic reference ID field on a step, so that given the type and that ID it's possible to correlate that step easily to an action, integration or integration pattern.

jimmidyson:
> maybe there should just be a generic reference ID field on a step, so that given the type and that ID it's possible to correlate that step easily to an action, integration or integration pattern.

That's an option but personally would prefer to have nullable FK columns to relevant tables. That at least allows us to enforce integrity in the DB.

jimmidyson:
Added tags.

gashcrumb:
> That's an option but personally would prefer to have nullable FK columns to relevant tables. That at least allows us to enforce integrity in the DB.

Yeah, that's fine.  One other thing, for steps, how would the choice EIP work with just a `step_index` field?  Or any other EIP that introduces a branch in the flow?

jimmidyson:
> how would the choice EIP work with just a step_index field? Or any other EIP that introduces a branch in the flow?

Hmm that is a good point. There is also properties of course... But I guess in that case they should each point to integration_step too? OK so let's discuss that in another PR when we need to :)

gashcrumb:
@jimmidyson sounds good.  For `integrations` I wonder if we could have a field the UI could use to persist a bit of state like what page the user was looking at etc, we might need to stick a stringified json blob in there for example.

jimmidyson:
I really don't want to persist UI state in DB - that just sounds wrong to me. But let's discuss it on a future PR :)

docs/database/schema.sql
```diff
@@ -51,15 +51,16 @@ CREATE UNIQUE INDEX connector_property_connector_id_property_index_uindex ON con
 -- "twitter")
 CREATE TABLE action
 (
-    id SERIAL PRIMARY KEY NOT NULL,
+    id INTEGER DEFAULT nextval('action_id_seq'::regclass) PRIMARY KEY NOT NULL,
     connector_id INTEGER NOT NULL,
     name TEXT NOT NULL,
     description TEXT,
     tags TEXT[],
+    camel_connector_maven_gav TEXT NOT NULL,
```
rhuss: generated png still shows maven_gav (which confused me a bit first). Thats much better ;-). Maybe one could omit "maven", as "gav" is probably expressive enough.

docs/database/schema.sql
```diff
@@ -51,15 +51,16 @@ CREATE UNIQUE INDEX connector_property_connector_id_property_index_uindex ON con
...
     CONSTRAINT action_connector_id_fk FOREIGN KEY (connector_id) REFERENCES connector (id)
 );
 CREATE UNIQUE INDEX action_name_connector_id_uindex ON action (name, connector_id);
-CREATE INDEX action_tags_index ON action USING GIN(tags);
+CREATE INDEX action_tags_index ON action (tags);
```
rhuss: It's too early for gin anyway ;-P

KurtStam:
TBH :
1. I guess I fail to understand why we have to change how properties are stored. This is going to make it less flexible for the UI how they want to store the form data right? I have not not seen argument or requirement on why we shouldn't simply use a string representation. I'm assuming we are still sending a JSON string back to the UI?
2. I guess I also fail to understand why we are going down to the DB schema level, as it takes away from the discussion on how data flows between Camel and the UI. I was hoping that the UML diagrams I created do just that job were we can simply persist the objects model coming through the REST API into the persistence store. We could have a discussion on whether we want to have a persistence layer with its own persistence objects.
3. So IMO we skipped sign off on 2 and have gone into creating a schema and we're discussing details that are more implementation and performance details like what to index.
4. Why we would we try to keep things simple and try to use the terminology used in our docs and the UI in our object model.

OK that feels better. Had to get that out. Maybe its just me.

Have a great weekend

jimmidyson:
> OK that feels better. Had to get that out. Maybe its just me.

Good! It's not just you, we're all working through this to try to come to the right solution for now.

> I guess I fail to understand why we have to change how properties are stored. This is going to make it less flexible for the UI how they want to store the form data right?

Why so? It is also not up to the UI how they want to store form data. The UI just should just workw with view objects, totally agnostic, unaware and uncaring about how data is stored. All the UI cares about is the REST API.

> I have not not seen argument or requirement on why we shouldn't simply use a string representation.

I think it was considered by the group that storing as a single string made things harder to grok and reason about. Considering the JSON is simply a structured object that maps quite nicely to table rows/columns that seemed to work nicely.

> I'm assuming we are still sending a JSON string back to the UI?

Yes but that bears no relation on how we store the data.

> I guess I also fail to understand why we are going down to the DB schema level, as it takes away from the discussion on how data flows between Camel and the UI.

Nothing flows from Camel to the UI. Camel is a runtime implementation of the designed integrations, the UI does not talk to the running integrations in any way.

> I guess I also fail to understand why we are going down to the DB schema level,

Because we should not necessarily be storing things in a way that is only for the UI. As others have expressed, the DB does not have to represent things in a way that is a complete fit for the UI. The UI just needs to work with the REST API. The DB schema is important, again as others have expressed, hence the need to produce this.

> we skipped sign off on 2 and have gone into creating a schema and we're discussing details that are more implementation and performance details like what to index.

We didn't skip sign off on anything, this was discussed and suggested by others on the team to do. We haven't discussed performance details or indices in any detail, the ones added are simple common sense indices, foriegn keys, etc based on the designed schema.

> Why we would we try to keep things simple and try to use the terminology used in our docs and the UI in our object model.

I feel we are trying to keep things simple... Our docs are outdated based on work done before the restart of iPaaS development. There have been a lot of changes, hence all the design work that has had to be done to conform to the new flow (e.g. Actions).

rhuss:
Thanks @KurtStam. I can feel with you, and I also can confirm that I had a hard time to build my mental domain view. Also I think there are many ways to model this domain<sup>**[1](#footnote)**</sup>, all with its pro and cons. 

> _kurt_: I have not not seen argument or requirement on why we shouldn't simply use a string representation.
>> _jimmi_: think it was considered by the group that storing as a single string made things harder to grok and reason about. Considering the JSON is simply a structured object that maps quite nicely to table rows/columns that seemed to work nicely.

If you hand through the camel configuration format directly to the UI in an opaque way, you couple the UI directly to Camel. Which means if Camel decides to change the format you would have to update the UI, but not ipaas-rest. In fact it make much sense to have it the other way round: If the Camel connector format changes we should adapt ipaas-rest, without touching ipaas-ui.

If the UI requirements need some structural changes, then also `ipaas-rest` should be changed, not the Camel connector format. 

So ideally, ipaas-rest would present in its API to the UI all the meta data in a form that is easily consumable by the UI (like description labels, data types, required ?, read-only ?, secret ?,...., already directly available) and then adapt this for the Camel connectors in a translation step.  Requiring an extra JSON parse step for the UI to get to the meta data (which btw can also fail differently and then the error must be handled, too) seems to me as a needless complication.

It's also much easier then to implement a REST call e.g. for building up a citizen developers form by sending a request "give me the meta data of all visible fields (i.e. not marked as hidden)" if this information is directly searchable in the backend.

This is all is assuming that we are talking here about the meta data originating from the Camel connector's [properties](https://github.com/redhat-ipaas/connectors/blob/master/connectors/http-connector/src/main/resources/camel-connector-schema.json#L21-L24) and the _configuredProperties_ which is the meta data massaged  by the expert dev.

For the actual _values_ I think we already agreed that this should be string-string key-values pairs (as this simple format is also expected by the Camel connector instances and will not change)

@KurtStam Does these arguments somewhat qualify as arguments why we shouldn't simply use a string representation for the meta data ?

----

<a name="footnote">1</a>: The model is simple but has three layers
* initial meta-data from camel
* massaged meta-data and prepared values from expert user
* remaining values from citizen user

rhuss:
@jimmidyson One feature of @KurtStam's approach is that the initial meta-data as defined by the Camel connectors is modelled within the domain (the `Connector` and `CamelConnector`). For the tables here the initial content for `connector`, `connector_property`, `action` and `action_property` must be initialised from the outside when a connector is created before they then can be individually changed. This can be done in an application setup / migration step.

That means that you can have for e.g. a "twitter" connector we have only one connector meta data (which fields are available with labels, hidden flag, ...). For the pre-filled values e.g. for credentials there can be multiple entries in `configured_connector` per `connector`. I think this is ok so, but want just to be sure that there is not supposed to be two different kind of _meta data per connector type_ sometimes which then can be individually changed by the expert dev (this would be covered by @KurtStam model).

davsclaus:
I am too blinded by Camel and would just confuse the model with my views. So let me try not to step on your toes.

At runtime all Camel just needs is a key/value pairs of the options to use for the Camel connector (is really a Camel component under the covers).

KurtStam:
@rhuss  @jimmidyson Yes that all makes perfect sense. It's just not the discussion I wanted to see. 

I'd would have liked to see a diagram with the data flow that marries rest-api objects to concepts from the requirements much like the format of PR https://github.com/redhat-ipaas/ipaas-rest/pull/120. It'd be nice to get some consensus as to what we are trying to build and to put up a rough architecture. 

my 2 cents

jimmidyson:
Speaking to @kcbabo, the Connection (configured_connector in this schema) is defined by the expert developer and is not editable/configurable by the citizen developer when they're creating an integration. That means that the current schema in this PR should work fine: when creating a Connection the expert developer will see the same fields for the same Connector, e.g. every Twitter Connection will show the same configuration fields when creating the Connection.

Hopefully I've explained that well enough - if there are doubts please raise them now. Otherwise please can we merge this ASAP so we can work on DB, model, API, etc as required?

gashcrumb:
@jimmidyson just to be clear, you mean this page, to the left, those fields, right?

https://redhat.invisionapp.com/share/9E9OFJDX3#/screens

jimmidyson:
@gashcrumb According to @kcbabo's response to my question about that on a call half an hour ago, I'm not sure what that page would be used for at all... The `Create Connection` page should be used to create a Connection (`configured_connector` in this schema), not to configure fields that can be filled in by a citizen developer. It is the definition of a Connection that a citizen developer can select when creating an integration.

gashcrumb:
@jimmidyson I suppose then that page is just for the expert user to set up the connection, i.e. he picks the fields to be configured, fills 'em in and we're good to go.

So then I'd suspect this page is really going to be anything specific to the action, i.e. say a filter for twitter-mention

https://redhat.invisionapp.com/share/HW9OF54BQ#/screens

jimmidyson:
@gashcrumb Yeah that sounds right to me. Action configuration is used in the Integration definition and needs to be filled in by cit dev and shouldn't include any Connection configuration.

gashcrumb:
+1 then :-)

KurtStam:
@jimmidyson, so can we have a nice writeup of the whole of it?

jimmidyson:
Give it another hour then I'll merge this if there's no more feedback.

@KurtStam @iocanel @rhuss @kahboom Any concerns please shout loudly and quickly ;)

jimmidyson:
@KurtStam If you have time, that would be great. Unfortunately I don't have time to do that right now as there's a whole heap of stuff still to do. I don't think there's anything too opaque in this model.

kahboom:
@KurtStam I'm working on a few diagrams and text for the website which might help, but obviously have been behind in just about everything. If you'd like to do a document write-up to help guide that, I'd be open to using that as the basis for the diagram and text?

@jimmidyson So, we are not going to allow expert users to define a connection with pre-filled values that would apply to all, for instance, Twitter services? (ie credentials) I know it does make things more complicated, such as if you wanted to use someone else's credentials and so on, but I just wanted to clarify before our UXD meeting.

jimmidyson:
> So, we are not going to allow expert users to define a connection with pre-filled values that would apply to all, for instance, Twitter services?

That is the definition of a Connection which is done by the expert user. They would create a Connection that would be used in Integrations. But those details wouldn't be visible to the cit dev when creating the Integration.

kahboom:
>That is the definition of a Connection which is done by the expert user. They would create a Connection that would be used in Integrations. But those details wouldn't be visible to the cit dev when creating the Integration.

Great, that's what my understanding was as well. I guess we can discuss how users can select certain configurations that are app-specific via the UI in today's UXD meeting (a feature implemented possibly in the future?).

My confusion was that we were expecting the design for Create a Connection to then be somewhat reflective of Stan's comment [here](https://github.com/redhat-ipaas/ipaas-rest/pull/119#issuecomment-283343415), although frankly I'd rather it look more like a pre-defined, non-optional set of fields without a side menu, drag-and-drop, etc. and that it be pretty generic (of course specific to the service). I'm slightly off topic here sooo... I'm okay with merging; my questions should be addressed somewhere else or in the meeting. :)

jimmidyson:
> I guess we can discuss how users can select certain configurations that are app-specific via the UI

They don't. They just select the Connection to use and then configure the Action accordingly.

kahboom:
>They don't. They just select the Connection to use and then configure the Action accordingly.
That was my original thought, but I think someone mentioned it / it was discussed that users might want to use another colleague's credentials/configuration settings, but in retrospect there is no strong use case for that.

jimmidyson:
They are effectively using their colleague's credentials, etc by using the Connection in their Integration... I'm missing something? But I don't think we need to worry about it and if we do, then we update it then.

jimmidyson:
[merge]ing

kahboom:
>They are effectively using their colleague's credentials, etc by using the Connection in their Integration... I'm missing something? But I don't think we need to worry about it and if we do, then we update it then.

Nope, that's correct. Apologies, I confused you with my confusion.

PR #120, opened 2017-02-23
UML for Connector, Connection and Integration

KurtStam:
I did some renaming and I think now Concepts, UI and Camel are well represented. Please review the UML diagrams along with the description of the data flow.

/cc @rhuss @iocanel @jimmidyson  @kahboom @gashcrumb


jimmidyson:
I'd like to consolidate this with #119 if possible. I see we can use concepts from both this and #119 to come to the right solution.

Some comments:

* We (@gashcrumb, @rhuss) discussed on IRC and decided that a single string for properties is not very configurable or usable. In #119 you can see that we separate out properties required for Connections and Actions.
* Having configurable fields on a Connection level means that a user has to configure those things when they create a Connection. This seems to be more a Connector level config, at least for now. While configurable properties on a Connection is more flexible, it also makes it harder to create a Connection, whereas I see the hard work in creating the Connector, not the Connection. That also means that Action is tied to a Connector rather than a Connection, which saves on duplication in the database.
* If previous point is accepted, then I don't see the need for CamelConnector type, that is all included in the Action, as we've done in #119.
* Camel connectors now include global config properties so we can use that for Connector properties and rest of properties from connector as Action properties.
* Are Actions usable as Steps (@gashcrumb, @kahboom?) or are they just start/end flows? Remember we can have child flows in future so Actions can be used as start/end in those child flows too.
* We decided for now to punt on IntegrationPatterns and hardcode step functionality via the UI as logic/config/etc for each step type will be custom to that step type. We can see how this works out and eventually would like to make it data driven but not right now.

So what I propose is to use #119 and include the Integration definitions very similar to this proposal, but with properties out in separate table.

gashcrumb:
On Feb 24, 2017 7:03 AM, "Jimmi Dyson" <notifications@github.com> wrote:

I'd like to consolidate this with #119
<https://github.com/redhat-ipaas/ipaas-rest/pull/119> if possible. I see we
can use concepts from both this and #119
<https://github.com/redhat-ipaas/ipaas-rest/pull/119> to come to the right
solution.

Some comments:

   - We (@gashcrumb <https://github.com/gashcrumb>, @rhuss
   <https://github.com/rhuss>) discussed on IRC and decided that a single
   string for properties is not very configurable or usable. In #119
   <https://github.com/redhat-ipaas/ipaas-rest/pull/119> you can see that
   we separate out properties required for Connections and Actions.
   - Having configurable fields on a Connection level means that a user has
   to configure those things when they create a Connection. This seems to be
   more a Connector level config, at least for now. While configurable
   properties on a Connection is more flexible, it also makes it harder to
   create a Connection, whereas I see the hard work in creating the Connector,
   not the Connection. That also means that Action is tied to a Connector
   rather than a Connection, which saves on duplication in the database.
   - If previous point is accepted, then I don't see the need for
   CamelConnector type, that is all included in the Action, as we've done in
   #119 <https://github.com/redhat-ipaas/ipaas-rest/pull/119>.
   - Camel connectors now include global config properties so we can use
   that for Connector properties and rest of properties from connector as
   Action properties.
   - Are Actions usable as Steps (@gashcrumb <https://github.com/gashcrumb>,
   @kahboom <https://github.com/kahboom>?) or are they just start/end
   flows? Remember we can have child flows in future so Actions can be used as
   start/end in those child flows too.

I wouldn't think an action would be a separate thing from the connection in
a flow, but part of the connection, with a step in the flow that refers to
the connection.



   - We decided for now to punt on IntegrationPatterns and hardcode step
   functionality via the UI as logic/config/etc for each step type will be
   custom to that step type. We can see how this works out and eventually
   would like to make it data driven but not right now.

So what I propose is to use #119
<https://github.com/redhat-ipaas/ipaas-rest/pull/119> and include the
Integration definitions very similar to this proposal, but with properties
out in separate table.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/redhat-ipaas/ipaas-rest/pull/120#issuecomment-282275878>,
or mute the thread

rhuss:
@KurtStam I have some questions in order understand you model better and to compare it to @jimmidyson approach:

* From a camel connector YAML you create entries as `Connector` (twitter) and `CamelConnector` (twitter-mention) which are not changed and server as meta data. It's an "import" of camel YAML into a DB representation and stays constant (until the next import). Correct ?
* Then there are `Actions` and `Connections` which are _instances_ of a `CamelConnector` and `Connector`, respectively (btw would love to see this reflected in the naming, too).  They also carry the configuration values + adaptable meta data, whereas the former (CamelConnector, Connector) are only for the metadata. Right ?
* Why do you fan out all `Actions` when a `Connection` is created ? Why not create an Action only when the user has selected a specific `CamelConnector` in the UI ? The other actions generated prematurely are probably never used, or misunderstand something here ?
* `Connection.configuredProperties` hold the meta data, which can be manipulated by an expert user but also the values which are 'sealed' and hidden (like credentials). Shouldn't be there a map holding those values, too ?
* Same for Action, there should be somewhere a place where the values can be stored. If I understand right, then the "Action"and its referenced "Connection"  is the full configuration (i.e. including the values entered by the users) which is sufficient to run a Camel connector. Is this correct ?

I think your model is not so far away from @jimmidyson suggestion, let me to try a mapping in the next comment (also for pointing out the differences as I see them). But first: lunch ;-)

rhuss:
What I think the biggest difference between #119 and #120 is that there is mismatch how meta data is stored (and what _connector_ means in both scenarios). 

Let me try to explain:

* In #119 the possible values for connectors (like 'twitter', 'salesforce') come from the outside and as soon a user has selected such a type an entry in `connector` is created on which an expert can work to change/massage the meta data.
* In #120 this possible values come from `Connector` and `CamelConnector` (so a #119 `connector` is **not** a #120 `Connector` but more a `Connection`, a meta-data editable connector, which groups actions)
* A #119 `action` is a #120 `Action`, but is only created when a user selected a specific action (which again the possible values come from outside the DB, in #120 its from the `CamelConnector`)
* In #120 there is currently no real place where to put the actual configuration _values_, i.e. the values a user as entered (not the meta data). They could be attached to `Action` and `Connection`, though, whereas in #119 they are stored in separate tables (`configured_connector` and `action_instance` from the alternative variations).

What I like about #119 is that it create 1 object per UI step. One could of course move the fixed background meta-data (i.e. the meta data coming from the Camel connector specs) also in DB tables, but that's probably not the point here.

But at the end both approaches are not so much different.

rhuss:
@jimmidyson @KurtStam , some additional comment on https://github.com/redhat-ipaas/ipaas-rest/pull/120#issuecomment-282275878

> Having configurable fields on a Connection level means that a user has to configure those things when they create a Connection. This seems to be more a Connector level config, at least for now. While configurable properties on a Connection is more flexible, it also makes it harder to create a Connection, whereas I see the hard work in creating the Connector, not the Connection. That also means that Action is tied to a Connector rather than a Connection, which saves on duplication in the database.

I think there is a mismatch as the `Connection` in #120 matches a #119 `connector` (see my comments above), so should be fine here (but please correct me, when I got it wrong)

> If previous point is accepted, then I don't see the need for CamelConnector type, that is all included in the Action, as we've done in #119.

The `Action` is an initial copy of a `CamelConnector` with the difference that it can be edited, whereas the `CamelConnector` data is _background data_, reused for every new similar `Action` created. @KurtStam please correct me when I'm wrong.

Except for the question from where to get the _background, constant camel connector meta-data _ which are used to initialise entries in _connector_ and _action_, and the difference that in #120 all actions (whether then used or not) are generated upfront I actually see both approaches quite similar. 

However, my urgent desire is that we get our naming straight and that we settle soon on a common language. TBH I have (still) a hard time to get the entity names in #120 into a mental picture, mostly also because the terms are so much overloaded. At least we have to resolve the "Connector" semantics. Would like to stick to the meaning of #119 here because most of us has already settled on this.


P.S. a quick shot for alternative names in #120 could be `CamelGlobalConfigurationMetaData`, `CamelActionConfigurationMetaData` as the types, and `Connector` + `Action` as the instances of this meta data where the meta data  be changed, and finally a `ActionInstance` which holds the configuration values. But thats just shooting from the hip.

jimmidyson:
I'm going to close this PR so we can continue the discussion on #119. Please comment/review on that PR when you get a chance.

PR #121, opened 2017-02-24
Use codacy for code coverage

jimmidyson:


circle.yml
```diff
@@ -13,36 +13,16 @@
...
-  - ./mvnw -Pfabric8 install
-  post:
-  - mkdir -p $CIRCLE_TEST_REPORTS/junit/
-  - find . -type f -regex ".*/target/.*-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
-  - cp runtime/target/classes/static/index.html $CIRCLE_ARTIFACTS/apidocs.html
-  - cp runtime/target/classes/static/swagger.json $CIRCLE_ARTIFACTS/
-  - cp runtime/target/classes/static/swagger.yaml $CIRCLE_ARTIFACTS/
-  - bash <(curl -s https://codecov.io/bash)
-
-deployment:
```
rhuss: The deployment step is not needed anymore ?

code-reports/pom.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rhuss: Couldn't this be also done directly in the top-level pom with a profile ?

Or what other purpose does this module serve ?

circle.yml
jimmidyson: This was an experiment...

code-reports/pom.xml
jimmidyson: No jacoco can only aggregate modules built before it in the reactor so can't be in the parent.

PR #122, opened 2017-02-24
Action Verifier Endpoint

rhuss:
Two API calls are proposed:

* Selecting an already existing connection (via id) and verifying it against the backend
* Taking a connector id and a set of options to verify the connectivity to a backend without actually creating an action. This should be used before an actual connection is created.

A "scope" is what should be validate and maps to the concept as described in https://issues.apache.org/jira/browse/CAMEL-10795 for the upstream verification logic.

This is also a good example where view objects (defined in DTO) are useful. A VConnectionConfiguration ("V" for Vendetta ... ehm ... View to distinguish from domain model objects) contains not a full connection but only a reference to the connector + the connector configuration data (+ maybe the action data). Introduced a package "dto" for that (which should stay in the rest package whereas the others eventually should move into own modules).

The code has been only roughly tested yet and open for discussion. As the model is still in flux it needs also to be decided on which handle to put the verify methods.

Related to IPAAS-203 (https://issues.jboss.org/browse/IPAAS-203)


davsclaus:
Okay so the ipass backend has this verifier api which then later on will find out how to call the camel ping check api (verifier) which Luca works on currently to be added to Camel 2.19.

rhuss:
@davsclaus Yes, that was the idea, or we even could short circuit directly in the UI without getting out to the REST API since I'm not sure whether we can merge this PR soon as the underlying stuff in the API is still moving. The net result will be the same, though: A "Test Connection" button which when pressed says: "All fine ;-)"

In the real implementation, there should be a dedicated external "Verifier Pool Manager" which is called synchronously via REST from this endpoint. This external Microservice's duty then is to manage a pool of 'Verifier Pods' each having access to one (or more) specific connector classes to trigger the real check. The pool is necessary to have already warmed up Pods so that we can save the startup costs.

These verifier Pods have also a REST API which in turn instantiates the connector and performs the check. There might different verifier pods for different connectors, it's the task of the "Verifier Pool Manager" to dispatch.

So the overall flow is expected to be something like: 
**iPaaS UI -> iPaaS REST -> Verifier Pool Manager -> Verifier Pod** 
(all services btw are stateless and independently scalable)

Currently everything is planned a synchronous access (which might lead to a blocking UI, though), but could be made asynrchronous on the set "iPaaS UI -> iPaaS REST" (or maybe later on as well)

rhuss:
@davsclaus The representation of the verifier result in iPaaS REST is highly inspired by the current work from @lburgazzoli at https://github.com/lburgazzoli/apache-camel/commit/d21c1e685c9e1e7bb370472a4b6cd99c75cadfae and we should keep that as aligned as possible.

davsclaus:
Sound good Roland. The Camel ping check may block while trying to connect, for example a HTTP server may be slow to respond, or not available on the network or something. There is no guarantee it can respond within a few seconds.

rhuss:
@davsclaus @lburgazzoli It would be awesome if the check itself could take some parameters in addition to the parameters required for the verification. That way one could add a timeout only for the verification, so the verify can return in a defined amount of time. This _processing instruction_ should be independent from the component implementation (so the available keys should be defined in the API), but a component might to decide to ignore a PI (e.g. when a timeout does not make sense for a particular component).

@lburgazzoli, what do you think ? Would it make sense to add an extra (optional) map argument to `ComponentVerfier.verify(...)` for this kind of _processing instructions_ ?

lburgazzoli:
@rhuss we certainly can support custom options, maybe with a prefix in the same map (i.e. camel.verifier.xyz as this is also a way used to customize some underlying libraries) but an option would also be to wrap it with a Future/CompletableFuture @ iPaaS side or eventually have a default method in the ConponentVerifier interface like asyncVerify

chirino:
looks like this needs a rebase

rhuss:
@chirino thanks for the heads up, will update on friday when I'm back.

iocanel:
retest this please

iocanel:
retest this please

chirino:
PR #223 extends this one, so I think it's safe to close this PR.

iocanel:
retest this

rhuss:
yes, can be closed.

PR #124, opened 2017-02-25
Add a postgresql JDBC driver

chirino:


KurtStam:
Hi @chirino I think it would make sense to add the connection info to a k8s secret instead. Also the file mount path could be by convention something like /kubernetes/secrets/ipaas-rest/postgres, so that if you create that file on your local machine, and add your local info it would run outside the container during development.

chirino:
You can just have the openshift templates mount the secret into the following env vars:
SPRING_DATASOURCE_USERNAME
SPRING_DATASOURCE_PASSWORD

and spring will override the default values with what it finds in the env var.

KurtStam:
And what about the connection info? Are there ENV Params for those too?

On Feb 27, 2017, at 19:05, Hiram Chirino <notifications@github.com> wrote:

You can just have the openshift templates mount the secret into the
following env vars:
SPRING_DATASOURCE_USERNAME
SPRING_DATASOURCE_PASSWORD

and spring will override the default values with what it finds in the env
var.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/redhat-ipaas/ipaas-rest/pull/124#issuecomment-282898567>,
or mute the thread

chirino:
Yes, you can configure just about  all spring application properties via environment  variables.

jimmidyson:
[merge]

PR #125, opened 2017-02-27
Add more tests for the event stream endpoints.

chirino:


runtime/pom.xml
```diff
@@ -378,11 +378,26 @@
...
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>com.launchdarkly</groupId>
       <artifactId>okhttp-eventsource</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp-ws</artifactId>
```
jimmidyson: this isn't required for okhttp 3.6 iirc

chirino: If I comment the okhttp-ws out, then I get a runtime exception in the test that looks like:

    java.lang.InstantiationError: okhttp3.internal.ws.RealWebSocket

jimmidyson: Ah this is because kube client is still on okhttp 3.4. We should look at upgrading that some time soon. This dependency stuff is feeling pretty fragile.

PR #128, opened 2017-02-28
Host the WebSocket endpoint under /api/v1/events.ws/*

chirino:
In short none of the SB web stuff can co-exist under /api/v1 as that’s 
being handled by resteasy.  Changed the impl to use a Undertow handler
similar to the SSE impl.  The two are more closely aligned now.


jimmidyson:
Thanks! [merge]

PR #138, opened 2017-03-06
Fixes #137. Throw the result of launderThrowable

### LINKED ISSUE #137: launderThrowable usage is inconsistent
In [launderThrowable](https://github.com/rhuss/ipaas-rest/blob/adc756b7b4b0f2f94b3664592968af730c776ed3/rest/src/main/java/com/redhat/ipaas/rest/v1/controller/handler/exception/IPaasServerException.java#L49-L49) there is a return value in certain occassion, however it is sometimes [ignored]( https://github.com/rhuss/ipaas-rest/blob/381e3a761be5732813a1026ab42ac55d9b831376/rest/src/main/java/com/redhat/ipaas/rest/v1/dao/DataManager.java#L120-L120).

Also I would either always throw an exception with this method having a void return value or always return an exception which then needs to be rethrown. I would prefer first, as this is less dangerous to forget it (although the IDE will probably croak because of a missing throw in the catch clause.

chirino:


jimmidyson:
[merge]

PR #141, opened 2017-03-07
Add Action REST API implementation

jimmidyson:
This adds handlers for `/actions`, `/actions/{id}`, `/connectors/{cid}/actions` and `/connectors/{cid}/actions/{id}`

I've also tidied up a few other minor things.

Ping @gashcrumb, @ryordan


rest/src/main/java/com/redhat/ipaas/rest/v1/handler/connection/ConnectorHandler.java
```diff
@@ -38,4 +39,9 @@ public String resourceKind() {
         return Connector.KIND;
     }
 
+    @Path("/{id}/actions")
+    public ConnectorActionHandler getActions(@PathParam("id") String connectorId) {
```
rhuss: Didn't know that you can return a handler. Again a noob question: From where does JAX RS know that it shouldn't use the handler as response object but look into for further processing ?

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/BaseHandler.java
```diff
@@ -25,6 +25,13 @@
     @Autowired
     private DataManager dataMgr;
 
+    public BaseHandler() {
+    }
+
+    public BaseHandler(DataManager dataMgr) {
```
rhuss: Wouldn't it be better to use constructor injection here (as we do in many other places, too ?). This would require to change subclasses to call the parent constructor is probably more straight forward.

If you don't want to bother on this level of details, I'm happy with it for now ;-)

rest/src/main/java/com/redhat/ipaas/rest/v1/util/ResultFilter.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Generic filter that removes items that do not match predicate
+ */
+public class ResultFilter<T> implements Function<ListResult<T>, ListResult<T>> {
```
rhuss: Would call it `PredicateFilter` since `ResultFilter` is a bit unspecific (the `PaginationFilter` is also a `ResultFilter`)

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/connection/ConnectorActionHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        return Action.class;
+    }
+
+    @Override
+    public String resourceKind() {
+        return Action.KIND;
+    }
+
+    @Override
+    public Action get(String id) {
```
rhuss: Ah, ok. Got it now, the annotations are on the super class. was a bit confusing for me.

Would it harm to repeat the annotations here (or maybe a simple comment would do it, too, for confused souls like me)

model/src/main/java/com/redhat/ipaas/model/integration/Step.java
```diff
@@ -32,7 +33,7 @@ default String getKind() {
         return KIND;
     }
 
-    IntegrationPattern getIntegrationPattern();
```
jimmidyson: Not right now, no. Not sure it ever will be tbh...

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/connection/ConnectorHandler.java
jimmidyson: Magic :) Honestly I don't know _how_ it works, I only learned this morning that this is the way to do it.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/BaseHandler.java
jimmidyson: Yeah I much prefer constructor injection where reference doesn't change during lifetime. Let me change that.

rest/src/main/java/com/redhat/ipaas/rest/v1/util/ResultFilter.java
jimmidyson: Renaming...

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/BaseHandler.java
jimmidyson: Done.

rest/src/main/java/com/redhat/ipaas/rest/v1/util/ResultFilter.java
jimmidyson: Done.

jimmidyson:
Let's [merge] this and improve stuff as needed later.

jimmidyson:
Hmm let's [merge] again - not sure what happened there.

rhuss:
@jimmidyson Could it be that a keycloak server is still running ? Locally when my integration test failed, the key cloak server has not been stopped sometimes.

jimmidyson:
[merge] more time for luck... otherwise manual merge.

PR #143, opened 2017-03-07
Ensure JsonDB orders by path

jimmidyson:
Otherwise DB records were not consistently allocated to correct object...

### DIFF
--- a/jsondb/src/main/java/com/redhat/ipaas/jsondb/impl/SqlJsonDB.java
+++ b/jsondb/src/main/java/com/redhat/ipaas/jsondb/impl/SqlJsonDB.java
@@ -116,7 +116,7 @@ public class SqlJsonDB implements JsonDB {
         Consumer<OutputStream> result = null;
         final Handle h = dbi.open();
         try {
-            ResultIterator<JsonRecord> iterator = h.createQuery("select path,value,kind from jsondb where path LIKE :like")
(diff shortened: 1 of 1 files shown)


jimmidyson:
retest this please

jimmidyson:
@chirino Happy for me to merge? This causes an issue if DB returns records not ordered by path (closes and opens items so you get multiple items with same path prefix).

jsondb/src/main/java/com/redhat/ipaas/jsondb/JsonDB.java
```diff
@@ -135,7 +135,7 @@ default String push(String path, byte[] json) {
 
     ///////////////////////////////////////////////////////////////////
     //
-    // Methods for working with JSON as Streams.. These methods provide
+    // Methods for working with JSON as Streams.. These methods provide`
```
rhuss: easter egg :-)

jimmidyson:
Ta! [merge]

jsondb/src/main/java/com/redhat/ipaas/jsondb/JsonDB.java
jimmidyson: Ha yeah i tidied that up before merge :-b

chirino:
Yep. Looks good I had assumed db would auto sort by PK.

PR #146, opened 2017-03-08
[WIP] Add a GitHub service

rhuss:
and hook it into IntgegrationHandler REST endpoint

This is still work in progress as it breaks currently the integration tests
because they try to reach out to GitHub for creating the repositories.

### DIFF
--- a/github/src/main/java/com/redhat/ipaas/github/GitHubService.java
+++ b/github/src/main/java/com/redhat/ipaas/github/GitHubService.java
@@ -45,8 +45,9 @@ public interface GitHubService {
      * Create or update file in a given repo on the fly.
      *
      * @param repo to update
+     * @param message commit message to use
      * @param files map of files with the keys being relative paths within the the repo
      *              and the values is the content in bytes.
      */
-    void createOrUpdate(String repo, Map<String, byte[]> files);
+    void createOrUpdate(String repo, String message, Map<String, byte[]> files) throws IOException;
 }
--- a/github/src/main/java/com/redhat/ipaas/github/GitHubService.java
+++ b/github/src/main/java/com/redhat/ipaas/github/GitHubService.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2017 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.redhat.ipaas.github;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import org.eclipse.egit.github.core.Repository;
+import org.eclipse.egit.github.core.client.GitHubClient;
+import org.eclipse.egit.github.core.service.RepositoryService;
+import org.springframework.stereotype.Service;
+
+@Service
+public class GitHubService {
+
+    private final RepositoryService repoService;
+
+    public GitHubService() {
+        // Maybe make the service name configurable ?
+        GitHubClient client = new GitHubClient("ipaas-github-proxy");
+        repoService = new RepositoryService(client);
+    }
+
+    /**
+     * Ensure that a given repository name exists. If it does not exist,
+     * it will be created as a public repository
+     *
+     * @param name name of the repository to create. This name should
+     *             be already a valid repo name. Must not be null.
+     */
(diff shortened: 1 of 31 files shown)


github/src/main/java/com/redhat/ipaas/github/GitHubService.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+public class GitHubService {
+
+    private final RepositoryService repoService;
+
+    public GitHubService() {
+        // Maybe make the service name configurable ?
+        GitHubClient client = new GitHubClient("ipaas-github-proxy");
```
jimmidyson: Do you not need the scheme in there? Just hostname?

rhuss: just hostname, according to the example at https://github.com/eclipse/egit-github/tree/master/org.eclipse.egit.github.core#using-github-enterprise

Still need to make a full turnaround ;-)

rhuss:
Integration should be working now. Next step: testing end2end :)

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -30,8 +34,11 @@
 @Component
 public class IntegrationHandler extends BaseHandler implements Lister<Integration>, Getter<Integration>, Creator<Integration>, Deleter<Integration>, Updater<Integration> {
 
-    public IntegrationHandler(@Autowired DataManager dataMgr) {
+    private GitHubService gitHubService;
+
+    public IntegrationHandler(@Autowired DataManager dataMgr, @Autowired GitHubService gitHubService) {
```
jimmidyson: I think we can remove these @Autowired annotations here btw.

github/src/main/java/com/redhat/ipaas/github/extended/FileContent.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package com.redhat.ipaas.github.extended;
+
+import java.util.Base64;
+
+/**
+ * Used for JSON serialization of a file content entry
+ */
+class FileContent {
```
jimmidyson: Would prefer at least fine getters! Assume this is using private fields for json serialisation right now? confused me!

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
rhuss: yes, makes sense. will remove them.

github/src/main/java/com/redhat/ipaas/github/extended/FileContent.java
rhuss: right, will add them

github/src/main/java/com/redhat/ipaas/github/GitHubServiceImpl.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private RepositoryService repoService;
+    private ExtendedContentsService contentsService;
+
+    @Value("github.service")
+    private String gitHubHost;
+
+    @PostConstruct
+    public void init() {
+        GitHubClient client = new GitHubClient(gitHubHost != null ? gitHubHost : "ipaas-github-proxy");
```
jimmidyson: How does this client use the keycloak token from the request?

rhuss: Very good point ;-) Haven't thought about this yet.

Suggestion is to create an extended version of the client for each request, adding the keycloak token. I guess it is contained in an header of the original request ? (which one ?).

will take care about that tomorrow.

rhuss: Ok, its the `Authorization:` header, as usual.

rhuss: Would it make sense to put the token into a ThreadLocal via a filter ? Otherwise it really clutters the API as it then has to be extracted individually in each handler and then passed to the GitHub Service for each call.

We could add a `AuthTokenHolder` class to the shared packages which is accessible from the other services. If we want to continue with the proxy approach, guess that's an easy way to make the token available globally.

Any thought on this ? Since its within a filter you also can remove the ThreadLocal directly in the same code where you have created it in a finally block, so there is no danger of leaking.

jimmidyson: You should be able to do something like:

```java
Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
if(authentication == null) {
    throw new IllegalStateException("Cannot set authorization header because there is no authenticated principal");
} else if (!KeycloakAuthenticationToken.class.isAssignableFrom(authentication.getClass())) {
    throw new IllegalStateException(String.format("Cannot set authorization header because Authentication is of type %s but %s is required", new Object[]{authentication.getClass(), KeycloakAuthenticationToken.class}));
} else {
    KeycloakAuthenticationToken token = (KeycloakAuthenticationToken) authentication;
    return token.getAccount().getKeycloakSecurityContext().getTokenString();
}
```

jimmidyson:
Merging... Thanks @rhuss!

PR #160, opened 2017-03-10
Ensure BC has correct git clone URL

jimmidyson:


rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -70,33 +70,29 @@ public String resourceKind() {
...
     }
 
     // ==========================================================================
 
-    private void ensureGitHubSetup(Integration integration, String secret) {
+    private String ensureGitHubSetup(Integration integration, String secret) {
         try {
             Integration integrationWithGitRepoName = ensureGitRepoName(integration);
-            String repoName = integration.getGitRepo().orElseThrow(() -> new IllegalArgumentException("Missing git repo in integration"));
+            String repoName = integrationWithGitRepoName.getGitRepo().orElseThrow(() -> new IllegalArgumentException("Missing git repo in integration"));
```
rhuss: Just a reminder: We should later be careful what to put into `Integration.gitRepo` as here is supposed the name and not the URL. Also suggest to change `gitRepo` to `gitRepoUrl`. 

For the name we can extract it from the URL anyway (should then be done in `ensureGitRepoName`.

ok for demo-time, will change that then later today.

jimmidyson: Agreed, this is demoware time sadly.

PR #167, opened 2017-03-10
Optimization for GitHub service

rhuss:
Instead of fetching all repositories and iterate over it (1 API call), fetch now the current user and fetch a repository directly (2 API calls).

It seems that this variant is not much faster, but more predictable as it not depends on the number of repoistories given.


rhuss:
retest this please

jimmidyson:
Last comment before i really go...

Jenkins PRs are bust like i said in my email to ipaas dev because Apache are blocking our access to snapshots repo that we need for camel snapshots... No idea why, we need to contact them to find out.

So have to rely on circle for now. Could possibly host own camel snapshots on bintray as well, manually building as required.

chirino:
retest this please

chirino:
[merge][lgtm]

PR #170, opened 2017-03-12
Add missing license headers.

iocanel:


rhuss:
we should try to fix the builds as @jimmidyson mentioned in https://github.com/redhat-ipaas/ipaas-rest/pull/167#issuecomment-285849829 so that we can start to make proper merge builds again.

iocanel:
[merge]

PR #171, opened 2017-03-12
Update the test model to use obejcts for action/connection properties.

chirino:


dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
```diff
@@ -111,7 +111,12 @@ public void store(ModelData modelData) {
             if (!id.isPresent()) {
                 LOGGER.warn("Cannot load entity from file since it's missing an id: " + modelData.toJson());
             } else {
-                if (fetch(kind, id.get()) == null) {
+                WithId prev = null;
+                try {
+                    prev = fetch(kind, id.get());
+                } catch (RuntimeException e) {
```
rhuss: Why catching and ignoring ? A comment would be nice, explaining why this would make sense here.

integration-to-maven-project-converter/src/main/java/com/redhat/ipaas/project/converter/DefaultIntegrationToProjectConverter.java
```diff
@@ -128,11 +129,11 @@ public DefaultIntegrationToProjectConverter(ConnectorCatalog connectorCatalog) {
...
 
                 try {
-                    ObjectNode stepNode = (ObjectNode) OBJECT_MAPPER.readTree(step.getConfiguredProperties());
-                    stepNode.set("kind", new TextNode(step.getStepKind()));
-                    if (step.getStepKind().equals(StepKinds.LOG)) {
-                        flow.addStep(OBJECT_MAPPER.readValue(stepNode.toString(), Log.class));
-                    }
+                    HashMap<String, Object> stepJSON = new HashMap<>(step.getConfiguredProperties());
+                    stepJSON.put("kind", step.getStepKind());
+                    String json = Json.mapper().writeValueAsString(stepJSON);
```
rhuss: If OBJECT_MAPPER is not used anymore, maybe change everytin to Json.mapper(). At least there should be no mixed usage (and eventually the constant OBJECT_MAPPER removed)

dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
chirino: I was thinking that we could get a error if the rest object can't be fetched due to a schema change.  But honestly, I think we would need to do a delete in catch to try to reset it properly.

integration-to-maven-project-converter/src/main/java/com/redhat/ipaas/project/converter/DefaultIntegrationToProjectConverter.java
chirino: Agree.

chirino:
Squashed and rebased to fixup conflicts with master.

model/src/main/java/com/redhat/ipaas/model/connection/Connection.java
```diff
@@ -49,7 +50,7 @@ default String getKind() {
 
     Optional<String> getConnectorId();
 
-    String getConfiguredProperties();
+    Map<String, String> getConfiguredProperties();
```
gashcrumb: Should the value really just be a string?  Isn't that a bit limiting given that we've support in the form config for numbers, booleans, etc?

chirino: At the end of the the day, it comes form an textual input box, and camel encodes it into a URL string.  I could open it up, but sounds like thats adding some complexity without much value.

chirino:
retest this please.

model/src/main/java/com/redhat/ipaas/model/connection/Connection.java
gashcrumb: Sounds good to me!

We can keep the form config as-is though with type support as that'll provide a better experience in the UI.

chirino: yep

cunningt:
retest this please.

cunningt:
retest this please.

chirino:
Going to manual merge this in as @gashcrumb's merge in  redhat-ipaas/ipaas-ui  depends on this going in.

PR #181, opened 2017-03-15
Change to using the most recent Fuse Camel 2.19 version.  Apache site is     timing out during use of 2.19 SNAPSHOT.

chirino:


chirino:
Yay.. finally build that passes all the tests!

PR #192, opened 2017-03-16
Cleanup deployment.json a bit, remove dup properties in the actions that existed in the connector properties, and sort the json properties around so that the objects are easier to identify (put kind first, then id, then name etc.)

chirino:


chirino:
retest this please

PR #194, opened 2017-03-16
Enabeling reading Enums as Strings from JSON

KurtStam:
This is so we can activate an integration by setting the StatusType to "Activated" on the integration.
When deployment starts it then sets the StatusPhase to Pending, and when up to Running, when not up to Failed, when successfully deactivated to Succeeded. There is also an Unknown Phase just in case. We could add a StatusMessage to report errors.


chirino:
You figured out the enum bits nice!

PR #201, opened 2017-03-17
Fixes #199

### LINKED ISSUE #199: How to update existing object?
Currently the UI tries to do a PUT but that fails at the OPTIONS request (snipped out all the javascript gobbledegook for ya):

```
OPTIONS https://ipaas-staging.b6ff.rh-idev.openshiftapps.com/api/v1/integrations/3 403 (Forbidden)

save-or-add-step:1 XMLHttpRequest cannot load https://ipaas-staging.b6ff.rh-idev.openshiftapps.com/api/v1/integrations/3. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:4200' is therefore not allowed access. The response had HTTP status code 403.
```

The response back from this OPTIONS request doesn't have any CORS headers, but we're sending:

```
Access-Control-Request-Headers:authorization,content-type
Access-Control-Request-Method:PUT
```

We should be using a PUT to update, right?

chirino:


KurtStam:
@chirino Do we need OPTIONS too?

PR #202, opened 2017-03-17
Adding Async version of our SimpleEventBus

KurtStam:
I think we do need an Async version of the EventBus to keep the UI responsive when we try to use 'Create Integration' events for things like creating the GitHub repo. 

Comments welcome


chirino:
This implementation could cause events to be delivered out of order.  I think it's better if you just implement subscribers that don't block.  You could put the threading logic on the GitHub side of things.

KurtStam:
@chirino I updated to run of a FIFO queue now, I hope takes care of your concern. The only issue I see is if a subscriber does something that blocks forever as that would stop the task thread.

KurtStam:
OK so we're not going to use the EventBus for this. So there is no need to make this async then.

PR #206, opened 2017-03-21
Add test support endpoints to aid e2e testing.

chirino:


chirino:
I need to rework this a bit..

PR #207, opened 2017-03-21
Change to auto-configuration of modules

### LINKED ISSUE #163: Change modules from component-scan to SB auto configuration
Similar how its done in the connector-catalog module (and removed from `Application`)

There should be an 'enable' flag for each module + a dummy implementation if enabled == false.

rhuss:
via META-INF/spring.factories . Fixes #163


rhuss:
[merge]

PR #208, opened 2017-03-21
Module hygienics

rhuss:
* "runtime" must only depend on "rest" (the one and only entrypoint) and the DAO implementations ("jsondb" for now). All other deps are deps of "rest" and are inherited transitively.
* "rest" must not depend on "jsondb" (or any other DAO implementation). "rest" and the DAO implementation are brought together by the "runtime" module as glue. The jsondb rest resource has been moved to the "jsondb" module.
* Rename "integration-to-maven-project-converter" to "project-generator" because that's it: A generator for a project (based on an integration but that's a detail). Also its more than Maven but contains the code, too. Main reason: Module name is much shorter ;-)


rhuss:
@chirino @KurtStam could you please do review asap ? because its some structural (though trivial) change and might lead to conflicts if merged in late (but with other PRs pending).

thanks ...

KurtStam:
I would prefer to rename it to integration-to-maven-project-converter-now-known-as-project-generator tbh.

chirino:
[merge]

Issue #142, opened 2017-03-07
Migration to new domain model


rhuss:
As we agreed somewhat on the new domain model based on this [DB schema](https://github.com/redhat-ipaas/ipaas-rest/blob/master/docs/database/schema.png) but still want to support view specific response objects (e.g. like a `Connector` with all its `Action`s in one case, only `Connector`s with no `Action` references in other cases) to remove efforts from the client side, I suggest the following rough migration strategy for the next sprint:

* The `model` module currently holds the "old" domain model, whose objects are directly returned to the client. The content of this model moves to the `rest` module and serves as the initial view objects. This means, that the client in this first step needs not to be changed.
* The domain model from #135 is moved into the `model` module.
* A custom mapping from view to domain needs to be added in the `rest` module. As the concepts are (hopefully) quite similar, this shouldn't be that hard. If we detect that it would be easier to change one of the view objects instead of a complicate mapping and if the client would benefit from this (because of a better consumable structure), we should then adapt the VOs and client instead.
* An additional `jpa` module can provide an alternative implementation of the interfaces in `dao`. This should be straight forward.

The advantage of this approach is that we can adapt the view layer incrementally for the client without doing a big bang where everything needs to be changed synchronously.

Does this makes sense or are there better alternative approaches ?

Comments are highly welcome ;-)

KurtStam:
OK I will a PR for the first two items
1. move whatever is in `model` back to rest.
2. use the `model` module to park the jpa entities.

Does that make sense? This shouldn't break anything.

rhuss:
This would work technically right now, but would introduce a dependency from dao -> rest as 
we for the moment would still use the old domain objects in the DAO layer.
This is then also a circular dependency (as we obviously have a dep from rest -> dao).

But you could introduce a "model2" and a "jpa" where we later rename "model2" to "model" and move the content of model to "rest".

So both approaches would work, guess its not a big problem to have the bogus deps for a temporary time.

The more difficult part will be the mapping between view objects and domain objects, which should happen in "rest", too.

KurtStam:
Model2 was introduced. We now need to refactoring to start using it.

rhuss:
@KurtStam imo i see the following tasks to do:

- [ ] Change *dao* to use *model2* object in their interfaces
- [ ] Adapt *jsondb* to fit these new dao interfaces
- [ ] Move the content of *model* as a subpackage of *rest*. The rest interfaces stay the same
[...]

PR #211, opened 2017-03-21
Add test support endpoints to aid e2e testing.

chirino:


chirino:
rebased

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/tests/TestSupportHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    public void resetDBToDefault() {
+        deleteAllDBEntities();
+        dataMgr.init();
+    }
+
+    @POST
+    @Path("/restore-db")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void restoreDB(ModelData[] data) {
+        deleteAllDBEntities();
```
rhuss: This works nicely for the moment because we expose domain models over the REST API but which we probably change later to using view objects (e.g. see #142) which are not always mapable 1-to-1 to domain objects. Of course we the can still adapt this later when we are at that point.

runtime/src/test/resources/application-test.yml
```diff
@@ -18,6 +18,10 @@ keycloak:
   enabled: true
   configurationFile: classpath:/ipaas-test-client.json
 
+endpoints:
```
rhuss: wonder whether we should think about introducing spring-boot profiles ? That way a profile can be easily selected during startup via an env (which we could add to the corresponding openshift template)

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/tests/TestSupportHandler.java
```diff
@@ -0,0 +1,81 @@
...
+@org.springframework.stereotype.Component
+@ConditionalOnProperty(value = "endpoints.test_support.enabled")
+public class TestSupportHandler {
+
+    private final DataManager dataMgr;
+    private final List<DataAccessObject> daos;
+
+    public TestSupportHandler(DataManager dataMgr, List<DataAccessObject> daos) {
+        this.dataMgr = dataMgr;
+        this.daos = daos.stream().filter(x -> !x.isReadOnly()).collect(Collectors.toList());
```
chirino: think that adding a ReadDataAccessObject might complicate the DataManager.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/tests/TestSupportHandler.java
chirino: agree.

runtime/src/test/resources/application-test.yml
chirino: Yeah not sure.

dao/src/main/java/com/redhat/ipaas/dao/init/ModelData.java
```diff
@@ -15,45 +15,73 @@
...
+    }
 
-	public void setKind(Kind kind) {
-		this.kind = kind;
-	}
+    @JsonRawValue
+    @JsonProperty("data")
+    public String getDataAsJson() throws JsonProcessingException {
+        if (json == null && data != null) {
+            json = data == null ? null : Json.mapper().writeValueAsString(data);
```
chirino: Should be fixed now.

dao/src/main/java/com/redhat/ipaas/dao/init/ModelData.java
```diff
@@ -15,45 +15,73 @@
...
+    }
 
-	@JsonRawValue
-	public String getData() {
-		return data == null ? null : data.toString();
-	}
+    @JsonRawValue
+    @JsonProperty("data")
+    public void setDataFromJson(JsonNode json) throws JsonProcessingException {
+        this.data = null;
```
chirino: Yeah.. lets avoid premature optimisations without hard data.

dao/src/main/java/com/redhat/ipaas/dao/init/ModelData.java
```diff
@@ -15,45 +15,73 @@
...
+    @JsonIgnore
+    public Object getData() throws IOException {
+        if (data == null && kind != null && json != null) {
+            data = Json.mapper().readValue(json, kind.getModelClass());
+        }
+        return data;
+    }
+
+    @JsonIgnore
+    public void setData(JsonNode data) {
```
chirino: should be fixed now

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/tests/TestSupportHandler.java
rhuss: I'm fine with it then. we can have a look later, its better to explain via code :)

chirino:
@rhuss 
I don't think we can use this as a general purpose backup ever since it does not stream the data, so you have to fit the whole DB in memory for these operations.

On the other hand, the jsondb endpoints do stream, so that could be used to backup the whole DB.

chirino:
squashed.

chirino:
[merge]

PR #215, opened 2017-03-21
Remove <from> image in the rest module so that fmp can select that ri…

iocanel:
Apparently the old assumption was not working correctly, so let's try a more direct approach.


iocanel:
This pull request lets fmp select the from image and the build strategy.

Locally, its solves the build issues I had and I hope it also fixes the image push issues in our rollout jobs.

PR #231, opened 2017-03-23
conforming to workflow statuses in PR #228

KurtStam:
Conforming to the Integration Workflow Designs from PR #228.


iocanel:
Merged #232 which seems to be almost a superset of this. So I am closing.

KurtStam:
Sure you can close, the only part on 232 I don't like is that Type and Phase are collapsed https://github.com/redhat-ipaas/ipaas-rest/pull/232/files#diff-1c27c75d9ac87ace25896541d0377031R34 
and he got rid of PENDING.

PR #232, opened 2017-03-23
Adding a controller module to manage transitioning integrations to their desired status.

chirino:


chirino:
I need to add some integration tests for this, but I think this implements the basic framework around managing the workflow state that @KurtStam, @iocanel and I discussed today.

model/src/main/java/com/redhat/ipaas/model/integration/Integration.java
```diff
@@ -31,9 +31,7 @@
 @JsonDeserialize(builder = Integration.Builder.class)
 public interface Integration extends WithId<Integration>, WithName, Serializable {
 
-    public static enum Type {Activated, Deactivated, Draft};
-
-    public static enum Phase {Pending, Running, Succeeded, Failed, Unknown};
+    public static enum Status { Draft, Activated, Deactivated, Pending, Deleted};
```
KurtStam: I don't think you ever desire for the state to be pending. I'd prefer to keep them separated.

controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java
```diff
@@ -0,0 +1,157 @@
...
+    private void enqueueCheckIntegration(String id) {
+        executor.execute(() -> {
+            checkIntegration(dataManager.fetch(Integration.class, id));
+        });
+    }
+
+    private void checkIntegration(Integration integration) {
+        Integration.Status desired = integration.getDesiredStatus().orElse(Integration.Status.Draft);
+        Integration.Status current = integration.getCurrentStatus().orElse(Integration.Status.Draft);
+        if (current != desired) {
```
KurtStam: want to pull in/rebase and add the deployment data from https://github.com/redhat-ipaas/ipaas-rest/pull/231?

KurtStam:
want to pull in/rebase and add the deployment data from https://github.com/redhat-ipaas/ipaas-rest/pull/231?

model/src/main/java/com/redhat/ipaas/model/integration/Integration.java
chirino: Removing the Pending status.  current state should be set to null where Pending was being used before.

controllers/src/main/java/com/redhat/ipaas/controllers/WorkflowHandler.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Set;
+
+/**
+ *
+ */
+public interface WorkflowHandler {
+
+    Set<Integration.Status> getTriggerStatuses();
+
+    Integration.Status execute(Integration model) throws Exception;
```
rhipaasuser: I would ditch the `throws Exception` as we wrap all exception under IPaaSException.

PR #236, opened 2017-03-27
Adding a SQL based LeaseManager impl so that we can distribute tasks across a cluster and make sure only 1 node process them at a time.

chirino:
Move the Recordings utility class to it’s own module.

### DIFF
--- a/leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
+++ b/leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
@@ -1,0 +1,293 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.redhat.ipaas.leases.impl;
+
+import com.redhat.ipaas.leases.Lease;
+import com.redhat.ipaas.leases.LeaseManager;
+import org.skife.jdbi.v2.DBI;
+import org.skife.jdbi.v2.Handle;
+import org.skife.jdbi.v2.PreparedBatch;
+import org.skife.jdbi.v2.util.LongColumnMapper;
+import org.skife.jdbi.v2.util.StringColumnMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import java.sql.SQLException;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.function.Consumer;
+
+@Component
+public class SqlLeaseManager implements LeaseManager {
+    private static final Logger LOG = LoggerFactory.getLogger(SqlLeaseManager.class);
+
+    private final DBI dbi;
+    private final ConcurrentHashMap<String, SqlLease> leases = new ConcurrentHashMap<>();
+    protected ExecutorService executor = Executors.newSingleThreadExecutor();
+    protected ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
+    private final String ownerId = UUID.randomUUID().toString();
+    private long leaseDuration = TimeUnit.SECONDS.toMillis(60);
+    private String insertSQL;
+
+    @Autowired
+    public SqlLeaseManager(DBI dbi) {
+        this.dbi = dbi;
+    }
+
+    @PostConstruct
+    public void start() {
+        executor = Executors.newSingleThreadExecutor();
+        scheduler = Executors.newSingleThreadScheduledExecutor();
+        withDB(dbi -> {
+            long dbTime = 0;
+
+            // The insert SQL statement used will depend on the DB type that we are using.
+            String dbName;
+            try {
+                dbName = dbi.getConnection().getMetaData().getDatabaseProductName();
+            } catch (Exception e) {
+                throw new IllegalStateException("Could not determine the database type", e);
+            }
+
+            if ("PostgreSQL".equals(dbName)) {
+                insertSQL =
+                        "INSERT into leases (lease, owner, expiration) " +
+                                "values (:id, :owner, :expiration) " +
+                                "ON CONFLICT(lease) DO NOTHING";
+
+                try {
+                    dbTime = dbi.createQuery("select extract(epoch from now()) * 1000")
+                            .map(LongColumnMapper.PRIMITIVE)
+                            .first().longValue();
+                } catch (Exception e) {
+                    // perhaps we are running agasint cockroachdb..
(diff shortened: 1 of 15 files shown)


zregvart:
retest this please

zregvart:
retest this please

jimmidyson:
add to whitelist

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @PostConstruct
+    public void start() {
+        executor = Executors.newSingleThreadExecutor();
+        scheduler = Executors.newSingleThreadScheduledExecutor();
+        withDB(dbi -> {
+            try {
+                long dbTime = 0;
+
+                // The insert SQL statement used will depend on the DB type that we are using.
```
KurtStam: Two things here:

1. Can we extract the DB specific SQL and put it in the config? This so the code stays DB independent. 
2. I would use the DB as a dumb store and put the logic and time handling in Java. Time is hard.. we have to make sure this works across DST etc. I think there is a hard requirement that time is synced on all nodes on kubernetes.

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
```diff
@@ -0,0 +1,283 @@
...
+    private void doLeaseMaintenance() {
+        withDB(dbi -> {
+            long now = currentTimeMillis();
+
+            // Delete expired leases..
+            dbi.createStatement("DELETE FROM leases where :now > expiration")
+                    .bind("now", now)
+                    .execute();
+
+            // Refresh our leases.
```
KurtStam: Where are the txn boundaries here? I would think there can be conflicts if the same code executes on different nodes with updates?

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
```diff
@@ -0,0 +1,283 @@
...
+            // lets find out which leases we now hold.
+            dbi.createQuery("SELECT lease FROM leases WHERE owner = :owner")
+                    .bind("owner", ownerId)
+                    .map(StringColumnMapper.INSTANCE).forEach(leaseId -> {
+                SqlLease lease = leases.get(leaseId);
+                if (lease != null) {
+                    aquiredLeases.remove(leaseId);
+                    lease.fireAquired();
+                } else {
+                    System.out.println("Weird we have a lease that we did not request");
```
KurtStam: logger?

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
```diff
@@ -0,0 +1,283 @@
...
+                // is too different from the DB time.
+                if (dbTime != 0) {
+                    long now = currentTimeMillis();
+                    long grace = leaseDuration / 4;
+                    if (!(now - grace < dbTime && dbTime < now + (leaseDuration / 4))) {
+                        throw new IllegalStateException("This machine's time is not synchronized with the database.");
+                    }
+                }
+
+            } catch (SQLException e) {
```
KurtStam: log error

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
```diff
@@ -0,0 +1,283 @@
...
+                            "INSERT into leases (lease, owner, expiration) " +
+                                    "values (:id, :owner, :expiration) " +
+                                    "ON CONFLICT(lease) DO NOTHING";
+
+                    try {
+                        dbTime = dbi.createQuery("select extract(epoch from now()) * 1000")
+                                .map(LongColumnMapper.PRIMITIVE)
+                                .first().longValue();
+                    } catch (Exception e) {
+                        // perhaps we are running agasint cockroachdb..
```
KurtStam: Not sure about this, but hopefully this goes away altogether if we use the db as a simple store.

test-support/src/main/java/com/redhat/ipaas/test/Recordings.java
```diff
@@ -13,18 +13,19 @@
...
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 
-import javax.annotation.Nonnull;
-
-import org.springframework.cglib.proxy.Enhancer;
-
 /**
  * This class allows you to proxy other objects and record all
  * method invocation done on those objects so that your test cases
```
KurtStam: neat :)

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
chirino: Putting DB specific config is painful for users as we are now pushing this hard task on to them.

The lease manager does generally use the DB as a dumb store, but it depends on all nodes int the cluster to have similar time which is why I added that bit of logic on startup to check to see if the node's  time is in sync with the DB node. We could remove that but then, you could run into cases where thing don't work right if a nodes time is way off.

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
chirino: which part are you not sure about?

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
chirino: I think each one of those SQL statements can be executed concurrently and multiple times without issue.  No transactions are needed. BONUS: less locking. 

Can you think of scenario where thats not the case?

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
chirino: will do.

leases/src/main/java/com/redhat/ipaas/leases/impl/SqlLeaseManager.java
chirino: will fix. thx!

rhuss:
Not yet diving in the code, but will it work with JPA ?

chirino:
I'm not sure JPA can do some of the SQL queries being used in this class. Why do you ask?

rhuss:
As said, I have not really yet a clear idea what this PR is supposed to do (not the PRs fault btw ;-), but want to know when using different, JPA based provider for the domain models (so not accessing the DB directly), that then this still works.

zregvart:
I wonder if PostgreSQL advisory locks would be good here (https://www.postgresql.org/docs/9.4/static/explicit-locking.html)

chirino:
@rhuss this does not depend on any of the domain models, so I think we are ok on that front.

chirino:
@zregvart I tried avoiding using features that can only be found in postgresql just in case we ever want to switch off it.

zregvart:
@chirino yeah, I've done that many times, it complicates the implementation and rarely does do the job well and in the end the need to switch to another backend never arises. Anyway just a thought since I started getting these on account of `retest this please` comment :)

iocanel:
retest this

KurtStam:
I'm still concerned about 

1. not having a PENDING state anymore so I don't really understand how the workflows (https://github.com/redhat-ipaas/ipaas-rest/tree/master/docs/workflows) work now and 
2. it still uses the time setting in the db rather then doing it in java code.

iocanel:
retest this please

iocanel:
retest this please

KurtStam:
@chirino @iocanel what is the status of this?

jimmidyson:
Is this still needed or can we close it now? If it's needed, then please rebase, fix merge conflicts and we can re-review? Thanks!

chirino:
closing.. I need to rebase and refresh this.

PR #243, opened 2017-03-27
Revert java base image to 1.2.3 (1.8.0_111)

### LINKED ISSUE #237: Exception thrown during boot when using DockerHub image
It seems that since release of `OpenJDK 1.8.0_121` image from DockerHub which is based on `fabric8/java-alpine-openjdk8-jdk/` there's an [exception thrown](https://gist.github.com/dsimansk/51f390d1e97b279deb347e9b04198017) during boot. I was able to reproduce it on Minishift and Dedicated env.

Googling around indicated that this may be caused by upgraded Java version. E.g. when I use f-m-p to build ipaas-rest with s2i binary (by default uses `fabric8/s2i-java` that is still on `_111`) the exception is gone.

rhuss:
This should fix #237

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -37,7 +37,7 @@
 
     <maven.build.timestamp.format>yyyy-MM-dd HH:mm</maven.build.timestamp.format>
 
-    <java-image.version>1.2</java-image.version>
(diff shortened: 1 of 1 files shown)


rhuss:
Going to merge that manually since there is not much to review :)

[merge]

PR #251, opened 2017-03-28
Verifier implementation using an external service (based on #122, supersedes #223)

rhuss:
This implementation calls out the external Microservice ipaas-verifier for doing the verification of a connector and has the following advantages over the locally process forking implementation of #223:

* Much faster because of an already 'prewarmed' running ipaas-verifier service
* Does not eat up memory and CPU resources by starting external Maven processes which are also quite network intensive since a lot of dependencies needs to be downloaded
* No local process management required
* Simpler implementation
* Doesn't need to call an external CLI command
* Standard REST / Microservice technology
* Supports a the full set of data that a Camel ComponentVerifier can return, in addition to `code` and `description`
  - Parameters which are affected by an error
  - Attributes holding additional error information (i.e. whether a parameter is missing or an unexpected parameter has been send.
 * Uses always both scopes: First verify the PARAMETERS before checking the CONNECTIVITY. Only if the initial parameter validation succeeds, a the real connection is called. This result in much faster responses if the given parameter set is unsound.
 * Does not add a GAV to the connector, which is factual _wrong_. GAVs are connected to Actions, a connector might reference an Action as its 'default verification action', but must not duplicate the (possibly contradicting) GAV of an action. Currently this association happens on the ipaas-verifier side. This had to be changed anyway.
 * Supports twitter, salesforce and http instead of only twitter (however could be easily added).
 
 The disadvantage of this approach is, that it is currently much simpler and supports three connectors: twitter, salesforce and http which are pre-started in the service. There is no dynamic loading of additional connectors, since this is not in the scope of the current Sprint. The setup is however expandable to deal with dynamic connectors, too.
 
 ---
 
The `LocalProcessVerifier` code is still present as reference, however due to changes in the model does not fully work and needs to be adapted (e.g. not to use a connector GAV but an action GAV and adapt to the interface of the Verifier). However I doubt that we should add efforts into this for now. Actually I would recommend to remove the code as it has served its purpose of a POC well but might not be needed anymore right now.
 
The templates has been adapted, the image builds has been setup and the Twitter validation has been tested end-to-end (starting from ipaas-rest, ipass-ui integration is still pending). For testing the Salesforce verification I would need proper credentials for it.


chirino:
Just want to make sure I highlight the long term problems that with running the verification code in process can encounter:
(1): The verification action code has to be compatible with all classes found in the classpath of the ipaas-verfifier.
(2): We need to make sure no system levels secretes are loaded into the ipaas-verfifier pod, as end user code will have access to it.
(3): We run the risk of an evil user defined connector stealing secrets that are being verified by another connector.

rhuss:
I completely agree with your remarks and we should carefully track them.

> (1): The verification action code has to be compatible with all classes found in the classpath of the ipaas-verfifier.

As briefly mentioned in the ipaas-verifier [README](https://github.com/redhat-ipaas/ipaas-verifier/blob/master/README.adoc#implementation-notes) the goal is to extract the hard coded calls into external, pooled pods which only run version of one (or multiple) connectors.

> (2): We need to make sure no system levels secretes are loaded into the ipaas-verfifier pod, as end user code will have access to it.

That's true (also for the forked approach since the filesystem is accessible here as well). Plus we need to ensure to secure the communication between ipaas-rest and ipaas-verifier (it's still http, although not exposed to the outside). For the final pod-pool approach however, the user code is run completely sandboxed within those pods with no secrets mounted.

> (3): We run the risk of an evil user defined connector stealing secrets that are being verified by another connector.

Not really sure if understand this right, but even now the credentials coming in are not stored anywhere but coming in as input params and used for the check statelessly. At least for the connectors that we provide, we are save. Custom connectors are supposed to run in external pods as sandboxes.

-----

The goal of this sprint is to provide a decent (== speedy) UX for validation of Twitter and Salesforce connections. We can add better isolation later if we target custom connectors (which is better than removing / changing stuff latter which nearly never happens if some time has passed)

@chirino Thanks a lot for the feedback! 'would be awesome if we could agree how to proceed today so that we could merge one of the PRs.

chirino:
In general I do like the idea of spinning a new pod per user to do validation as that isolates the damage he can cause, But I'm not sure what's the plan to deal with the classpath issue.  Perhaps long term we should move the local process validator logic into the ipaas-verifier, and any unblessed (connectors we are not sure if they are compatible or not) get validated via forking?

chirino:
Oh and by issue (3) I meant that I could craft a component which figures out how to install intercept subsequent HTTP requests to the verifier.  Perhaps via a servlet filter.

rhuss:
Yes, that could be an option, too. But I'm not have any idea how a custom connector should be realized at all. 

*  How does an expert user define a customer connector ? 
* How is the custom connector build and made available for the verifier ? 
* How will a custom connector be made available for the UI ? 
* How enters the custom connector the DB ? 
* How are the custom connectors separated for each tenant ? 

I think these are all question which should be tackled before thinking about how to execute such a custom connector. My gut feeling is that the answers to those questions will guide us in a certain direction we don't yet fully know. Let's do first things first.

jimmidyson:
I'm not too worried about (3): the SaaS solution will use only curated connectors, not user-provided. Self-managed deployments will be up to the user to ensure they review connectors for deployment.

I would prefer to go with this PR's approach right now. I haven't reviewed code - would appreciate it if @KurtStam or @iocanel could do so please.

The sooner we can get to the pooled pod approach the better, but this works for now.

chirino:
LGTM

rhuss:
[merge]

rhuss:
[merge]

PR #267, opened 2017-03-30
adding statusMessage

KurtStam:
Added the Integration.statusMessage back, and added a message for the Timed integration example. 

I also increased the db version so it will refresh the data on staging when merged and deployed.

### DIFF
--- a/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
+++ b/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
@@ -498,7 +498,8 @@
       "name": "Timed Pull to Post Example",
       "description": "This is an example of a Timed Pull to Post Integration",
       "configuration": "",
-      "desiredStatus": "Activated",
+      "desiredStatus": "Deactivated",
+      "statusMessage": "This integration failed. Please contact system admin for errors.",
       "tags": [
         { "name": "example" }
        ],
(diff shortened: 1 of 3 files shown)


kahboom:
I may not be a reviewer, but I strongly approve of this.

dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
```diff
@@ -498,7 +498,8 @@
       "name": "Timed Pull to Post Example",
       "description": "This is an example of a Timed Pull to Post Integration",
       "configuration": "",
-      "desiredStatus": "Activated",
+      "desiredStatus": "Deactivated",
```
kahboom: Yes! Now we can use `desiredStatus` as well.

kahboom: Which was available, but there was no point in using as none of the data used it. @gashcrumb WDYT?

gashcrumb: @kahboom we've actually setting `desiredStatus` when saving the integration depending on which button you use.  Recent change I made gets this state reflected in the UI though really the state that we should be showing is actually `currentStatus`, not `desiredStatus`.

KurtStam:
merge please

PR #269, opened 2017-03-31
Fix potential NPE.

iocanel:


model/src/main/java/com/redhat/ipaas/model/connection/Connector.java
```diff
@@ -72,7 +72,10 @@ default Connector withId(String id) {
     default Map<String, String> filterSecrets(Map<String, String> properties, Function<Map.Entry<String, String>, String> valueConverter) {
         return properties.entrySet()
             .stream()
-            .filter(e -> this.getProperties() != null && this.getProperties().containsKey(e.getKey()) && this.getProperties().get(e.getKey()).getSecret())
+            .filter(e -> this.getProperties() != null
+                && this.getProperties().containsKey(e.getKey())
+                && this.getProperties().get(e.getKey()).getSecret() != null
```
rhuss: Just a general question (without knowing the wider context): Could secret be a `boolean` instead of a `Boolean` ? what's the case for a null value for getSecret() ?

jimmidyson: Safer to use something like `Boolean.TRUE.equals(this.getProperties().get(e.getKey()).getSecret())` perhaps?

jimmidyson:
[merge]

PR #271, opened 2017-03-31
Add an offlineRefreshToken field to user and a lastUpdatedByUserId field to integration.

chirino:


KurtStam:
1. It seems not very secure to me to store this token in the integration, as it would then be send back to the UI. I think it makes more sense to track them in a separate object, and maybe a token table?

2. Is the requirement to track the complete edited-by history or only who touched it last? Would we remove the 'User' object, or is that going to be used for something else?

chirino:
stale

Issue #274, opened 2017-04-04
Pull out the 'kind' from the dataType field for data shape


gashcrumb:
Since we bake the` kind` into the dataType field (like `'java:foo.blah.Myclass'`) now I have to have special parsing client-side.  Why not have the `kind` in a separate field?

gashcrumb:
Another option is just align to the URL scheme that the data mapper uses in it's mapping file maybe, though the inputs for the component are just the class names for Java.

KurtStam:
@gashcrumb The metadata in Camel have it as one field. I like consistency :). I can break it apart if you really want. If so 'none' and '*' would go in the kind field I think.

gashcrumb:
hmm, I don't know.  I guess maybe we'll have to revisit this when we deal with `json` and other non-java cases.  I definitely think *not* passing along a URI-like syntax would be better in general,

jimmidyson:
I much prefer the separate field approach that @gashcrumb proposes. The `kind` can be the distinguishing field to parse to specific types as needed. Much easier to work with IMO.

KurtStam:
@gashcrumb PR is ready for merge

PR #276, opened 2017-04-04
issue #274

KurtStam:
Splitting DataShape.dataType up into kind and type

### DIFF
--- a/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
+++ b/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
@@ -249,7 +249,7 @@
              "kind" : "none"
           },
           "outputDataShape": {
-             "kind" : "*"
+             "kind" : "any"
           },
           "properties": {
             "httpUri": {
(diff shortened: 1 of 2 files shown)


KurtStam:
@gashcrumb Does this look out you?

gashcrumb:
@KurtStam +1 though what does `'*'` vs `'none'` mean I wonder...

jimmidyson:
@gashcrumb Perhaps `any` vs `none` would make more sense?

gashcrumb:
@jimmidyson yeah, that's more self-describing I think.

dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
```diff
@@ -244,10 +246,10 @@
           "name": "HTTP GET",
           "description": "Call a service that is internal (within your company) or external (on the internet) by specifying the service's URL",
           "inputDataShape": {
-             "dataType" : "none"
+             "kind" : "none"
           },
           "outputDataShape": {
-             "dataType" : "*"
+             "kind" : "*"
```
jimmidyson: Can we use `any` instead of `*` please?

KurtStam:
@gashcrumb 

`none` means there is not data going in/out. Like twitter there in no data going 'in', only 'out', so dataShapeIn is `none`. Would you prefer it not set?

`*` or soon to be `any`refers that it is free form. Like for the HTTP GET connector it can be anything.

gashcrumb:
let's leave it as `'none'` I think.

KurtStam:
@jimmidyson I think we are good to go on this. Can you approve & merge plz?

Issue #277, opened 2017-04-04
Make sure currentStatus is set on integrations


gashcrumb:
A few integrations don't have `currentStatus` set, and some do, the `currentStatus` field should be set on each integration.

As a workaround in the UI we're actually just showing `desiredStatus`, but I'd like to fix this => https://github.com/redhat-ipaas/ipaas-ui/issues/384

kahboom:
+1 This is a big deal, as we are planning on tracking "in progress" states (ie integrations in the process of being activated/deactivated), where integrations may be transitional and cannot be deleted or activated, etc. We also check both status types for different reasons, including discrepancies between them to determine whether or not to display certain things.

KurtStam:
This has to do with the [workflow](https://github.com/KurtStam/ipaas-rest/tree/master/docs/workflows) setting the status, which is what @iocanel is working on. Th example integrations have this field set just so that you the UI can display something. 

I'm currently setting the initial state to 'Draft' when the integration is first created (POST).

kahboom:
@KurtStam - Since @iocanel 's PR was merged do we have `currentStatus` being updated properly? We would need it to be set properly once integrations are created, and we can then set `desiredStatus` ourselves, check for discrepancies to determine an 'In Progress' state, etc.

KurtStam:
@kahboom I think it now is hooked up. It should follow the stati as described in the workflow diagrams:
https://github.com/KurtStam/ipaas-rest/tree/master/docs/workflows. Maybe want to double check with @iocanel

gashcrumb:
BTW, the `Twitter to Salesforce Example` integration doesn't return a currentStatus.  Also with my code in the linked PR all the integrations just show spinners since `currentStatus` != `desiredStatus.

kahboom:
>BTW, the Twitter to Salesforce Example integration doesn't return a currentStatus. Also with my code in the linked PR all the integrations just show spinners since currentStatus != `desiredStatus.

Yesterday all but Activated integrations were returning a currentStatus. Activated did not show a currentStatus. I wonder if it's related?

kahboom:
I'd need to confirm that though, not sure there is a direct correlation but one or two were definitely not even returning that property at all.

rhuss:
I just updated deployment.json it with #316 so I set the current and desired status to "Activated" for now. Will be available as soon as this PR is merged.

rhuss:
[...]

PR #279, opened 2017-04-04
issue #277, setting the initial status to Draft

KurtStam:


rhuss:
lgtm.

one minor thing: it would be awesome if PRs would not be named after issue numbers only but with a more descriptive name. (Not so much an issue with this PR but more as a general remark).

This would help tremendously in understanding what the Github notification mails are about by reading the subject only and without directly jumping to the browser. The issue number can still be referenced within the PR comment (which for non-trivial PRs would be awesome to have, too).

KurtStam:
+1 May I add to that that we should add the issue number on the commit comment.

PR #292, opened 2017-04-11
use cache before going to the db

KurtStam:


KurtStam:
@chirino @jimmidyson can I get a review on this?

Issue #305, opened 2017-04-13
Support for Tags


kahboom:
## Overview
Add support for Tags. NOTE: We do not have designs for a page in the UI to manage tags, so the user will be using something like select2 (see demo [here](http://valor-software.com/ng2-select/) and click the 'Multiple' tab), where the user can type in Tags while creating an Integration and press enter to save it.

## Requirements:
- A `/tags` endpoint to GET all tags. Should not need any other CRUD actions here, unless you want to knock them out ahead of time. Good in case users make a typo and have OCD. We don't want anyone having a conniption either.
- A method to load tags automatically for Integrations as an associated model. I don't think we will ever need to load Integrations without Tags, but if you think it should be separated out for less overhead and more flexibility, feel free to do it as a query param like `/integrations?tags=1`. Of course, let me know so that I can update the request in the UI.
- The UI avoid having to do a separate request for all tags (only if we need to for a dedicated Tags page, for instance) and then filtering through them, assigning them by ID to the proper Integration, etc. All of that should be done API-side.
- A method to load tags when we request a single Integration with `/integrations/:id`We can do this one of two ways (not sure how you are doing this atm):
  - `/integrations/:id/tags`
  - `/integrations/:id?tags=1`
- The Tag should allow spaces, numbers, and maybe certain special characters (will leave that up to you guys to decide, but consider use cases like `e2e` or `front-end`).
- The user should be able to POST a Tag when creating their Integration. You can either allow it as an array of strings or array of objects as part of the larger Integration object when we POST/PUT at the end of the creation work flow, or as a separate POST the moment they press Enter in the Tags input field (however you will not have the Integration ID it will be associated with yet as the Integration won't have been created; this may be trickier). I don't care either way.

## Up for Discussion / Lower Priority
Possibly for post-Summit, but some things to consider:

- A `/tags/:id/integrations` endpoint to GET all integrations associated with a particular Tag. This may be lower priority. Use case: the user clicks a Tag/label from an Integration Detail page and it will list all Integrations with that same tag (will not include Templates, or anything else with the same Tag). In the future, we can support `/tags/:id/` to load everything with that Tag, or `/tags/:id/templates` once we begin to incorporate Templates.
- Additional CRUD actions for the general `/tags` endpoint. I will leave it up to you to decide whether or not to allow anything other than GET here, because, as I stated earlier, we currently do not have a page to manage them, so any changes to tags will be done within the context of an Integration or Template.
[...]

PR #309, opened 2017-04-14
adding support for tagging connections and integrations, issue #305

KurtStam:
Addresses: https://github.com/redhat-ipaas/ipaas-rest/issues/305

@kahboom Connections and Integrations can now be tagged using the `tags` field. The `tags` field support a list of strings. Tags are always set on the connection or integration. You can obtain a list of  tags that are currently in use using GET /api/v1/tags/, which will return a Alphabetically ordered list of tags. So you can simply start using tags on connections and integration without predefining a list of tags. I think this should take care of all the requirements.  Your widget may need to add a `new tag` to create a new tag that is not yet in the list.

Let's get started with this and we discuss the lower priority list if needed.

Note that this also needs a database reset, but that will conflict with reset ordered in https://github.com/redhat-ipaas/ipaas-rest/pull/307. So please merge #307 first and I will add a commit to reset the db.


model/src/main/java/com/redhat/ipaas/model/WithTags.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package com.redhat.ipaas.model;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface WithTags {
+
+    Optional<List<String>> getTags();
```
jimmidyson: Should this be a set rather than a list? Semantically doesn't allow duplicates so makes more sense to me.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/tags/TagHandler.java
```diff
@@ -0,0 +1,69 @@
...
+        Set<String> tags = new TreeSet<String>();
+        addTags(tags, getDataManager().fetchAll(Integration.class));
+        addTags(tags, getDataManager().fetchAll(Connection.class));
+        ListResult<String> result = ListResult.of(tags);
+        return result;
+    }
+
+    private void addTags(Set<String> tags, ListResult<? extends WithTags> items) {
+        if (items.getItems()!=null) {
+            for (WithTags item : items.getItems()) {
```
jimmidyson: This looks fine but would read really well as stream - up to you if want to change it over, not that bothered.

jimmidyson:
Can we keep tags ordered when storing too please so UI doesn't have to order itself? Need consistency in tag order.

KurtStam:
@jimmidyson thx for the feedback. I have incorporated it and rebased. It should be good to go now.

KurtStam:
@jimmidyson I've incorporated the feedback and this should be good to go now.

model/src/main/java/com/redhat/ipaas/model/WithTags.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package com.redhat.ipaas.model;
+
+import java.util.Optional;
+import java.util.TreeSet;
+
+public interface WithTags {
+
+    Optional<TreeSet<String>> getTags();
```
jimmidyson: Why not just `Set`instead of `TreeSet`?

KurtStam: So it's alphabetically ordered.

rhuss: @KurtStam but then it should be `SortedSet` in this interface if you want to express the intend that it is a sorted set. Implementation classes like `TreeSet` or `HashSet` shouldn't be included in APIs.

KurtStam: @rhuss yes good point. Fixed.

jimmidyson: In this case, I would just expect to use Set and document sorted nature, remembering this is going to be exposed over the REST API.

jimmidyson:
retest this please

KurtStam:
retest this please

PR #311, opened 2017-04-17
IPAAS-314: Have an Online and Demo versions of the workflow handlers.

chirino:
The demo version move to the desired state after 2 seconds.

### DIFF
--- a/controllers/src/main/java/com/redhat/ipaas/controllers/NoopHandler.java
+++ b/controllers/src/main/java/com/redhat/ipaas/controllers/NoopHandler.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


controllers/src/main/java/com/redhat/ipaas/controllers/NoopHandler.java
```diff
@@ -0,0 +1,46 @@
...
+package com.redhat.ipaas.controllers;
+
+import com.redhat.ipaas.model.integration.Integration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.util.*;
+
+@Component
+@ConditionalOnProperty(value = "controllers.integration.enabled", havingValue = "noop")
```
KurtStam: controllers.integration.enabled can have values of true/false/noop?

chirino: thats it.

KurtStam: So what's the difference between noop and false? Maybe you can add some javadoc, and a paragraph to the readme?

chirino: noop does not update integrations, false basically enabled demo mode.  it update integration to the desired status after 5 seconds.  Naming might not be great but it's what's easiest to use with the current templates.

PR #314, opened 2017-04-18
Fixes #312 Avoid re-triggering workflow action every time an integration is updated.  Only do it if it’s desired state changes or if it’s not currently scheduled for an workflow action.

### LINKED ISSUE #312: IntegrationController event cycle
When an exception occurs while enqueue() in IntegrationHandler, the Integration object is updated, which causes another changed-event to be enqueue(), where the same exception occurs, ... and so on.

--> https://github.com/redhat-ipaas/ipaas-rest/blob/32c2d7fad1bd9092f61aad5b395d69df201a9889/controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java#L148-L157

chirino:


controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java
```diff
@@ -114,7 +117,9 @@ private void checkIntegration(Integration integration) {
         Optional<Integration.Status> current = integration.getCurrentStatus();
         if (desired.isPresent() && !current.equals(desired)) {
             WorkflowHandler workflowHandler = handlers.get(desired.get());
-            if (workflowHandler != null) {
+
+            String scheduledKey = "" + integration.getDesiredStatus() + ":" + integration.getId().get();
+            if (workflowHandler != null && !scheduledChecks.containsKey(scheduledKey)) {
```
rhuss: Could this check be put directly within enqueue() (at the beginning) so that it doesn't depend on the order the methods are called, and its 'selfcontained' (e.g. scheduledChecks is only known to and used by `enqueue()`. The key needs to be calculated only once, too.

controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java
```diff
@@ -114,7 +117,9 @@ private void checkIntegration(Integration integration) {
         Optional<Integration.Status> current = integration.getCurrentStatus();
         if (desired.isPresent() && !current.equals(desired)) {
             WorkflowHandler workflowHandler = handlers.get(desired.get());
-            if (workflowHandler != null) {
+
+            String scheduledKey = "" + integration.getDesiredStatus() + ":" + integration.getId().get();
```
rhuss: The integration id can be (and was) null. See e.g. https://github.com/rhuss/ipaas-rest/blob/c5b788abb7b2fc53514717bb3c61eae1192a7ffb/rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java#L120-L120 for a usage where the id is null.

controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java
chirino: sounds good.

controllers/src/main/java/com/redhat/ipaas/controllers/IntegrationController.java
```diff
@@ -125,7 +123,8 @@ private void checkIntegration(Integration integration) {
     private void enqueue(WorkflowHandler handler, String integrationId) {
         executor.submit(() -> {
             Integration integration = dataManager.fetch(Integration.class, integrationId);
-            if (stale(handler, integration)) {
+            String scheduledKey = "" + integration.getDesiredStatus() + ":" + integrationId;
```
rhuss: Not 100% sure about the key. Shouldn't it be also the currentStatus included ? E.g. when the desiredStatus stays the same but the current status changes, shouldn't it be then rechecked, too ?

could be that i'm talking bs here ;-)

chirino: I was thinking that we mainly want to re-trigger the workflow processing immediately if a user changes the desired status.  Since current status is only changed by workflow actions, I think it's ok if those subsequent workflow actions occur after the scheduling delay.

rhuss: ok fine. lets give it a try.

PR #316, opened 2017-04-19
Refactoring of Integration Status Change Event handlers

rhuss:
* Cleaned up interfaces a bit and reorganized packages
* Fixed integration update logic. (e.g. the controller is only responsible for updating the status
* Introduced a `StatusChangeHandlerProvider` interface for a single entry point (so that components can be switched more easily)
* Simplified Demo and Noop implementation
* Fixed Github file update issue by inserting a sleep
* Tuned logging
* Fixed `desiredStatus` of integrations in deployment.json


controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
```diff
@@ -74,13 +72,17 @@ public void start() {
...
-                    if ((changeEvent != null
-                        || changeEvent.getKind().isPresent() && changeEvent.getId().isPresent())
-                        && (Kind.from(changeEvent.getKind().get()) == Kind.Integration)) {
-                        enqueueCheckIntegration(changeEvent.getId().get());
+                    if (changeEvent != null) {
+                        changeEvent.getId().ifPresent(
+                            id -> changeEvent.getKind().ifPresent(
+                                kind -> {
+                                    if (Kind.from(kind) == Kind.Integration) {
+                                        enqueueCheckIntegration(id);
```
rhuss: @chirino I think its false to enqueue a check for each change on the integration as during processing the event the integration gets updated as well (like status change or adding the GitHub repo). So you end up with multiple jobs for the same integration which then all start a build.

See e.g this logs.

<details>

```
2017-04-19 08:28:14.564  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Draft) --> enqueuing
2017-04-19 08:28:14.566  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Processing integration 3.
2017-04-19 08:28:15.346  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:28:15.347  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:twitter-mention-connector:0.3.6
2017-04-19 08:28:48.919  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:28:48.919  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:28:48.921  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:salesforce-upsert-contact-connector:0.3.6
2017-04-19 08:29:00.679  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:29:00.679  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:30:00.680  INFO [-,,,] 1 --- [pool-5-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:30:00.681  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Processing integration 3.
2017-04-19 08:30:01.377  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:30:01.377  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:twitter-mention-connector:0.3.6
2017-04-19 08:30:01.388  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:30:01.388  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:30:01.388  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:salesforce-upsert-contact-connector:0.3.6
2017-04-19 08:30:12.274  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:30:12.274  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:31:12.274  INFO [-,,,] 1 --- [pool-5-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:31:12.275  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Processing integration 3.
2017-04-19 08:31:12.919  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:31:12.919  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:twitter-mention-connector:0.3.6
2017-04-19 08:31:12.930  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:31:12.930  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:31:12.930  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:salesforce-upsert-contact-connector:0.3.6
2017-04-19 08:31:22.462  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:31:22.463  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:32:22.463  INFO [-,,,] 1 --- [pool-5-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:32:22.464  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Processing integration 3.
2017-04-19 08:32:23.103  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:32:23.103  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:twitter-mention-connector:0.3.6
2017-04-19 08:32:23.113  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:32:23.113  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.p.c.DefaultProjectGenerator        : Adding step null [integration Optional[3]]
2017-04-19 08:32:23.114  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.catalog.ConnectorCatalog         : Downloading Maven GAV: com.redhat.ipaas:salesforce-upsert-contact-connector:0.3.6
2017-04-19 08:32:32.362  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:32:32.363  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:33:32.363  INFO [-,,,] 1 --- [pool-5-thread-1] c.r.i.c.i.IntegrationController          : Integration 3 is not in the desired status Activated (current: Pending) --> enqueuing
2017-04-19 08:33:32.364  INFO [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Processing integration 3.
2017-04-19 08:33:32.403 ERROR [-,,,] 1 --- [pool-4-thread-1] c.r.i.c.i.IntegrationController          : Error while processing integration status for integration 3

```
<details>

Any idea how we should resolved this without introducing a central object holding all actively running checks ? // @jimmidyson

rhuss: // @iocanel

rhuss:
@iocanel @jimmidyson Could I have one review please ? It's mostly about rearranging things (e.g. moving back OpenShift logic to the OpenShift service, etc.), but would love to have this is in (or rejected) soon to avoid rebasing marathons. Thanks ! 

Unfortunately with this sort of refactoring smaller PRs (which I would prefer of course) are not so easily possible.

connector-catalog/src/main/java/com/redhat/ipaas/connector/catalog/ConnectorCatalog.java
```diff
@@ -21,12 +21,16 @@
...
 import org.apache.camel.catalog.maven.MavenArtifactProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.URISyntaxException;
 import java.util.Map;
 
 public class ConnectorCatalog {
 
+    Logger log = LoggerFactory.getLogger(this.getClass());
```
jimmidyson: Should be private - thanks codacy!

controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
```diff
@@ -24,39 +31,29 @@
...
-import java.util.concurrent.*;
-
 /**
  * This class tracks changes to Integrations and attempts to process them so that
  * their current status matches their desired status.
  */
 @Service
 public class IntegrationController {
-    private static final Logger LOGGER = LoggerFactory.getLogger(IntegrationController.class.getName());
+    private final Logger log = LoggerFactory.getLogger(this.getClass());
```
jimmidyson: Nit: why prefer instance field rather than static?

controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
jimmidyson: I think the better way to do this is to have a reconciliation loop that reconciles desired vs actual state. This of course introduces a potential delay between requesting stuff and stuff actually happening, but I think that is acceptable as long as the loop interval is small enough. It reduces the requirement to know the state of running checks, plus allows failing over to a redundant controller without having to share state between processes.

rhuss: I think, too, that it makes sense to have a central and global loop and don't let the handlers themselves manage the scheduling. This would avoid situations likes this.

As this would be some major change (post summit ?), I suggest as a quick fix here to only react on changes in non monitored integrations. @chirino added already something for this, but it seems not to work yet properly.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
rhuss: just a habit which is useful when used in a base class. These are singletons anway so no much different in number of created logger objects. But I don't mind revert it to static.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
jimmidyson: OK agreed, let's go with a quick fix here. What code are you referring to that @chirino added to do this? Are there obvious problems with it we can quickly fix?

controllers/src/main/java/com/redhat/ipaas/controllers/integration/IntegrationController.java
jimmidyson: Don't worry about it. I prefer the instance field approach too, it allows for changing log level per instance, rather than at the logger name as per normal.

rhuss:
Fixed problem with multiple checks for the same integration

This is uses a state map which is checked when processing a 'change-event' so 
that no second job is triggered if there is already one.

Also added some logging and clarified the various steps happening (these could be even fine granular integration step states)

However there is still the problem with expiring tokens. Because of the 
reordering, a GitHub API will fail first, whose error message is a bit obfuscated (its a JSON deserialisation error coming from the egit lib)

Would be awesome if there would be a check for verifying whether a token is till valid. 

@iocanel @jimmidyson Any idea how such a check could look like ?

BTW, I still get the GitHub errors from time to time, even when waiting 1s between update over the API.

rhuss:
Another problem with out logic: When the final deployment fails on OpenShift because the created project has some issue (like it is these days), then the Integration stays in status "Pending" and ipaas-rest will constantly update the GitHub repo with the generated files (regardless whether they have been changed), recreate the OpenShift resources and start a (costly) S2I source build. There is no way yet to move to state "Failed" after some retries. 

Maybe when the source state is "Pending" one should simply skip all the initial setup step and only check for the Deployment to be there ? This can be easily added.

rhuss:
Last push for today. It's as good as it gets for now ;-) GitHub API is still somewhat flaky (e.g. needs also some time after repo creation before files can be added, sometimes 1s is not good enough ;-(.

I highly recommend that we should go ways from GitHub API. It seems to have some serious timing issues.

jimmidyson:
@rhuss The token itself contains validity info (it's a JWT). I can't remember exactly the best way to parse it out, but pretty sure keycloak provides something in their library. I can't look right now, but will be able to find something out I'm sure if you don't get anywhere with it.

rhuss:
@jdyson added a JWT expiry checks so that we can now nicely bail out when things time out. Of course, there is still the issue what todo when a token expires. Ideally it could be refreshed (or it could be an offline token from the very beginning, if supported). At least the loop should be stopped and the integration should be moved into a (not yet existing) error state. At the moment it repeatedly try to process the integration but will always fails because noone updates the token.

If this is ok so far, it would be awesome if we could merge asap (there are already other PR pending touching this code regions).

rhuss:
Added test for Token.isTokenExpired()

... and fixed an issue. That's what tests are for ;-)

BTW, would be awesome to add a test for valid token (with expiration == Integer.MAX_INT), but tbh don't know how to create and sign such a token.

rhuss:
retest please

PR #317, opened 2017-04-19
Use a custom settings.xml that uses our internal nexus as a mirror of…

iocanel:
… central and jcenter.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,8 +3,8 @@ def mavenVersion='3.3.9'
 slave {
     withOpenshift {
             //Comment out until pvc issues are resolved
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -75,14 +75,14 @@
   </distributionManagement>
 
   <repositories>
-    <repository>
+    <!--repository>
```
jimmidyson: Doesn't this mean local builds will fail?

iocanel: If bintray is used just as a central mirror, everything should be fine.

It seems though, that the swagger plugin depends on stuff that are exclusive to bintray, so we have to use it and provide an additional mirror for that in nexus.

iocanel: In other words, we need to uncomment that and make sure our nexus image provides a bintray proxy.

jimmidyson: Agreed. So should we close this PR then?

iocanel:
retest this please

Issue #318, opened 2017-04-19
Need more sample integrations in various states


gashcrumb:
@KurtStam to implement this view it'd be great to have a few more integrations that are fixed in these various states just so the view can be further developed:

![d4e20a08-1560-11e7-9b14-490f46a96169](https://cloud.githubusercontent.com/assets/351660/25183095/94ca397a-24e4-11e7-970b-4b4fb91c6342.jpg)

It doesn't really work to create the integrations from the UI and hope that they work or don't work, and anyways all the integrations disappear eventually.

@kahboom FYI

kahboom:
+1 This would be awesome.. some 'In Progress', some failed, some passed, some drafts, some inactive, etc. FYI the green checkmark means that there is no `statusMessage`, which would mean it has passed, however, the integration can still be a Draft or just Inactive, for example.

KurtStam:
@gashcrumb @kahboom Please review the PR to see if this has all the state combos you need.

PR #319, opened 2017-04-19
Need more sample integrations in various states #318

KurtStam:
A few more integrations that are fixed in various states just so the view can be further developed

### DIFF
--- a/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
+++ b/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
@@ -1694,7 +1694,7 @@
     {
     "kind": "integration",
     "data": {
-      "id": "6",
+      "id": "7",
       "name": "Deleted Client Service Example",
       "description": "This is an example in Deleted state",
       "configuration": "",
--- a/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
+++ b/dao/src/main/resources/com/redhat/ipaas/dao/deployment.json
@@ -1098,5 +1098,597 @@
         }
       ]
     }
+  },
+  {
+    "kind": "integration",
+    "data": {
+      "id": "3",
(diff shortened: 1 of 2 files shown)


KurtStam:
1. For as far as I understand it you can't set the desiredState to Pending, but the currentState can be pending while things are getting deployed etc.

2. I've added another example with a Deleted states.

rhuss:
> Pending is a state? When was this introduced, or maybe I need to regen our model then?

@gashcrumb The state "Pending" is used when a OpenShift build has started but the application is not yet deployed (its determined with the number of pods running. 0: pending, 1: activated). What I currently miss is an "Error" state, e.g. when the build fails or the pod canot be deployed (as it is the case at the moment because of some service account related issues).

However, when we introduce an "Error" state we have to think also how the user can deal with it:

* Showing some logs
* Restarting the step causing the error
* Deleting the integration
....

Otherwise its more frustrating than having an ever lasting "Pending" state, because there is still hope ;-)

rhuss:
@gashcrumb The current states are

```java
enum Status { Draft, Pending, Activated, Deactivated, Deleted};
```

gashcrumb:
@rhuss awesome, thanks for clarifying.  This is definitely a case where it would have been good to know this was added when it was added so we can run our swagger generator client-side.

@kahboom I'll run the generator and get this updated, we'll have to adjust the views, I guess now the spinner should just show up when the `integration.currentStatus` is at `Pending` state and we can get rid of the current logic that shows the spinner when there's a mismatch between `desiredStatus` and `currentStatus`

kahboom:
> Otherwise its more frustrating than having an ever lasting "Pending" state, because there is still hope ;-)

lol! This cracked me up. I think eventually an "Error" state would be convenient, but I agree that it's better to wait and do it properly. Thank you @rhuss and @KurtStam !

PR #320, opened 2017-04-20
Wire up the mapper to the project generation.

chirino:


project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
```diff
@@ -137,9 +136,11 @@ public DefaultProjectGenerator(ConnectorCatalog connectorCatalog, ProjectGenerat
     }
     */
 
-    private byte[] generateFlowYaml(GenerateProjectRequest request) throws JsonProcessingException {
+    private byte[] generateFlowYaml(Map<String, byte[]> contents, GenerateProjectRequest request) throws JsonProcessingException {
         Flow flow = new Flow();
+        int stepCounter[] = new int[]{0};
         request.getIntegration().getSteps().ifPresent(steps -> {
+            stepCounter[0]++;
```
rhuss: Would use a plain int here and declare it within the closure since it is used nowhere outside of the closure.

rhuss: Is this counter not only to be supposed to increased for "data-mapping" steps instead for every kind of step ?

project-generator/src/main/resources/com/redhat/ipaas/project/converter/templates/pom.xml.mustache
```diff
@@ -39,6 +39,17 @@
       </snapshots>
     </repository>
     <repository>
+      <id>sonatype.snapshots</id>
+      <name>Apache Development Snapshot Repository</name>
```
rhuss: "Sonaytpe Development Snapshot Repository"

project-generator/src/main/resources/com/redhat/ipaas/project/converter/templates/pom.xml.mustache
```diff
@@ -137,6 +148,13 @@
       <version>${funktion.version}</version>
     </dependency>
 
+    <!-- atlasmap runtime -->
+    <dependency>
+      <groupId>io.atlasmap</groupId>
+      <artifactId>camel-atlasmap</artifactId>
+      <version>1.10.0-SNAPSHOT</version>
```
rhuss: Please move versions into a parent pom's `<dependencyManagement>` section.

chirino: parent pom? this pom has no parent.

project-generator/src/main/resources/com/redhat/ipaas/project/converter/templates/pom.xml.mustache
chirino: good catch

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
```diff
@@ -161,6 +162,18 @@ public DefaultProjectGenerator(ConnectorCatalog connectorCatalog, ProjectGenerat
                     continue;
                 }
 
+                if ( "mapper".equals(step.getStepKind()) ) {
```
chirino: StepKinds.MAPPER does not exist, but I can make it a constant.

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
chirino: if done in the closure then it would reset or every step and not actually count up.

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
```diff
@@ -47,6 +45,7 @@
 
     private final static ObjectMapper YAML_OBJECT_MAPPER = YamlHelper.createYamlMapper();
     private final static String PLACEHOLDER_FORMAT = "{{%s}}";
+    private final static Charset UTF8 = Charset.forName("UTF-8");
```
chirino: ok.  I do wish the JDK had this as a constant somewhere since all JDKs are required to support this encoding.

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
rhuss: Still don't get it. Isn't it supposed that the counter is _within_ the loop over all steps ? afais (sorry, its a bit messed up in the diffs view, probably need to checkout) the counter is increased only once,  so always 1 used in the conditional below. But I have to check out the code to verify this ...

project-generator/src/main/resources/com/redhat/ipaas/project/converter/templates/pom.xml.mustache
rhuss: sorry, thought it was in the project.

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
```diff
@@ -162,6 +172,18 @@ public DefaultProjectGenerator(ConnectorCatalog connectorCatalog, ProjectGenerat
                     continue;
                 }
 
+                if ( MAPPER.equals(step.getStepKind()) ) {
```
rhuss: Seems that one of my last comments got lost (probably didn't submitted it and cant find it anymore): Actually instead of having here many conditionals I'd prefer some sort of 'handle' abstraction which is looked up via the step kind. This would make the code smaller and easier to understand. also you don't have this nasty loop control via `continue`

project-generator/src/main/java/com/redhat/ipaas/project/converter/DefaultProjectGenerator.java
rhuss: Couldn't we introduce a StepKinds.MAPPER (like there is a StepKinds.ENDPOINT) ? If the mapper has a special meaning and needs special treatment, then is should be treated as a special step, too. Actually I would even make that an enum.

PR #326, opened 2017-04-21
Make the statusMessage option in the StatusChangeHandler Configure the fabric8.mode for the fabric8 plugin.

chirino:


pom.xml
```diff
@@ -117,6 +117,7 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <fabric8.mode>kubernetes</fabric8.mode>
```
rhuss: I'd prefer to add it directly to the f-m-p configuration with 

```xml
<configuration>
    <mode>kubernetes</mode>
    ...
</configuration>
```

It's close where it is used. Can still be overwritten from the outside. 

Plus I'm not 100% sure whether it breaks something different, but I _think_ that the CI build use source s2i (so not f-m-p at all), but not 100% sure. If it used binary s2i then setting this mode would prevent it. @iocanel is my assumption correct and is it save to nail the mode down to Kubernetes?

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
```diff
@@ -37,26 +37,30 @@
 
         class StatusUpdate {
             private Optional<Integration.Status> newStatus;
-            private String message;
+            private Optional<String> message;
```
rhuss: wouldn't it makes sense to move this property to `errorMessage` as it is then clearer what it is used for (at least at the moment) ?

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
```diff
@@ -37,26 +37,30 @@
...
-                this(status, null);
+                this(Optional.of(status), Optional.empty());
             }
 
             public Optional<Integration.Status> getNewStatus() {
                 return newStatus;
             }
 
-            public String getStatusMessage() {
+            public Optional<String> getStatusMessage() {
```
rhuss: Don't you have to adapt the usage of this method, too ?

pom.xml
chirino: ok took this bit out of this PR.  I'll let you figure out how to best do this.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
chirino: Renamed to mach the fields in the integration.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
chirino: No, don't think so.

rhuss: ok, it's fed into an immutable builder which has overloaded methods for this.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
rhuss: Actually I don't meant the status object, think `newStatus` is  a good fit for a `StatusUpdate`. I meant to rename `message` --> `errorMessage` as it is supposed only be set in the error case.

rhuss: But 'status' is ok, too (just not so nice ;-)

PR #331, opened 2017-04-24
Move the schema check logic to it’s own service and use it to control when the deployment.json is imported.

chirino:
Caching logic was not exactly right for when the cache did not hold the entire dataset (like when we restart and the dataset is in the DB but not in the cache).

### DIFF
--- a/dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
+++ b/dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
@@ -75,7 +75,9 @@ public class DataManager implements DataAccessObjectRegistry {
         for (DataAccessObject dataAccessObject : dataAccessObjects) {
             registerDataAccessObject(dataAccessObject);
         }
+    }
 
+    public void resetDeploymentData() {
         if (dataFileName != null) {
             ReadApiClientData reader = new ReadApiClientData();
             try {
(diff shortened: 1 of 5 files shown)


dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
```diff
@@ -140,13 +142,14 @@ public String generatePK(Cache<String, WithId> entityMap) {
 
     @SuppressWarnings("unchecked")
     public <T extends WithId> ListResult<T> fetchAll(Class<T> model, Function<ListResult<T>, ListResult<T>>... operators) {
-        Kind kind = Kind.from(model);
```
jimmidyson: What are these changes about? Seem unrelated to the schema check/update service?

chirino: See commit message: 'Caching logic was not exactly right for when the cache did not hold the entire dataset (like when we restart and the dataset is in the DB but not in the cache).'

after restart fetchAll was returning back an empty list, since we were trying to only fetch from the cache.

jimmidyson: Having a PR that includes unrelated changes to the pr title, albeit with details in commit messages, doesn't ​make reviews easy... 

Once we get post demo we're going to work out better ways to do this.

jimmidyson:
@iocanel Build isn't running on rh-idev so no status check for PRs... Any idea what's going on?

dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
```diff
@@ -140,13 +142,14 @@ public String generatePK(Cache<String, WithId> entityMap) {
...
-        Cache<String, WithId> cache = caches.getCache(kind.getModelName());
 
-        ListResult<T> result = ListResult.of((Collection<T>) cache.values());
-         
-        if( result == null ) {
+        ListResult<T> result;
+        if( getDataAccessObject(model)!=null ) {
```
jimmidyson: Isn't this going to ignore the cache if there is a dao for this kind? Not sure that's what we want to do.

chirino: Right.  This PR breaks the build without the cache changes.

chirino: Well it only ignores the caches for fetch all.  Single lookups should be cached.  At scale this is what you would want anyways right?

dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
chirino: Its not unrelated.  The caching changes have to go otherwise the tests would be breaking.

dao/src/main/java/com/redhat/ipaas/dao/manager/DataManager.java
jimmidyson: Who knows what we want at scale? It all needs revisiting post-summit. I'd even consider a distributed cache that we use for persistence, potentially even obviating the need for PostgreSQL...

Regardless, right now we're not using the cache properly anyway so let's merge this and revisit it later.

jimmidyson:
Manually merging...

PR #335, opened 2017-04-25
The naive DataManager.generatePK method  breaks down when the cache does     not contain ALL entities.  Refactor the key generator in JsonDB to the core     module, and use that instead.

chirino:


jimmidyson:
Do you want to cherry pick this into summit-demo branch?

PR #345, opened 2017-04-26
Set status to "pending" when the desiredStatus changes.

### LINKED ISSUE #342: Saving an integration puts it in the Draft status first, then after a while it goes In Progress.
UI displayed state seems to be purely driven by the currentStatus.  Perhaps we should avoid putting into Draft initially.

### LINKED ISSUE #343: Wrong desired state when deactivating integration ?
[This](https://github.com/rhuss/ipaas-rest/blob/8d1d6030be55cea310e20f9cead70d7acbfd7149/rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java#L62-L62) looks so obvious wrong that I wonder whether it is intentional (and it was updated 5 days ago). If so, a comment would be nice ;-)

rhuss:
Fixes #342 and also fixes #343 which it is based on.

Normally would branch from master but this time its easier as I hope both PRs go in (and they overlap a bit).


rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -95,6 +92,17 @@ public Integration create(Integration integration) {
         return Creator.super.create(updatedIntegration);
     }
 
+    private void updateStatus(String id, Integration.Status desiredStatus) {
```
jimmidyson: This is better named `updateDesiredStatus`.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -95,6 +92,17 @@ public Integration create(Integration integration) {
...
+        this.update(id, new Integration.Builder()
+            .createFrom(get(id))
+            .desiredStatus(desiredStatus)
+            // Set the current status to 'pending' immediately when
+            // a status change is requested.
+            // This status will be later changed by the activation handlers.
+            .currentStatus(Integration.Status.Pending)
```
jimmidyson: I would actually expect the controller to do this when it notices that desiredStatus != actualStatus. The REST API shouldn't change current status IMO.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
rhuss: agreed

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
rhuss: Yeah, don't like the imperative style, too (as it's not the callers duty to do state changes)

let me think about it when I pick up changes from #349

model/src/main/java/com/redhat/ipaas/model/integration/Integration.java
```diff
@@ -65,6 +65,8 @@ default Kind getKind() {
 
     Optional<Status> getCurrentStatus();
 
+    Optional<List<String>> getStepsDone();
```
jimmidyson: I'm not keen on adding steps done to the integration, as this will be sent to the clients too, which I guess isn't so bad in hindsight. Once we review the API models though, I'd prefer to differentiate between spec (desired) and status (current) as is done in Kubernetes API. That simplifies things quite a lot, easier to read, and not conflate current and requested state. Don't worry about changing this now though. Full review post-summit.

rest/src/main/java/com/redhat/ipaas/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -92,11 +92,19 @@ public Integration create(Integration integration) {
         return Creator.super.create(updatedIntegration);
     }
 
-    private void updateStatus(String id, Integration.Status desiredStatus) {
+    private void updateDesiredStatus(String id, Integration.Status desiredStatus) {
         this.update(id, new Integration.Builder()
             .createFrom(get(id))
             .token(Tokens.getAuthenticationToken())
             .desiredStatus(desiredStatus)
+            // Set the current status to 'pending' immediately when
```
jimmidyson: Thanks for the comment - again this is fine for now, but we need a clearer separation of concerns going forward.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
```diff
@@ -36,33 +36,39 @@
...
+                this(status, null, stepsPerformed != null && stepsPerformed.size() > 0 ? stepsPerformed : null);
             }
 
-            public Optional<Integration.Status> getStatus() {
+            public Integration.Status getStatus() {
                 return status;
             }
 
-            public Optional<String> getStatusMessage() {
+            public String getStatusMessage() {
```
chirino: This should be an optional, otherwise your get NPES.
```
java.lang.NullPointerException: statusMessage
	at java.util.Objects.requireNonNull(Objects.java:228) ~[na:1.8.0_111]
	at com.redhat.ipaas.model.integration.ImmutableIntegration$Builder.statusMessage(ImmutableIntegration.java:2459) ~[model-0.1-SNAPSHOT.jar!/:na]
	at com.redhat.ipaas.controllers.integration.IntegrationController.lambda$callStatusChangeHandler$10(IntegrationController.java:179) ~[controllers-0.1-SNAPSHOT.jar!/:na]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_111]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[na:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) ~[na:1.8.0_111]
	at java.lang.Thread.run(Thread.java:745) ~[na:1.8.0_111]


```

model/src/main/java/com/redhat/ipaas/model/integration/Integration.java
rhuss: I think these should be two different objects, and the processing parts (like 'desiredStatus' and 'stepsDone') should go into a separate object.

jimmidyson: Do you mean a separately persisted object or a nested object? I was meaning nested objects, `integration.spec` and `integration.status` that define desired and current state respectively so you'd have `integration.status.currentStatus` and `integration.status.stepsDone` (ignore naming) to describe current state. We will look at this next week, post-summit so I wouldn't spend too much time thinking about this particular PR, if it gets past the immediate issue.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
rhuss: My ide warns me to use Optional<> as fields or even method parameters. And there are people backing this opinion:

* http://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments
* http://dolszewski.com/java/java-8-optional-use-cases/
... and so on

personally, I think we should not overuse it, so it's hopefully ok to fix this where it is used. Its a helper multi-valued return value anyway.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/online/ActivateHandler.java
```diff
@@ -40,6 +40,10 @@
 
 public class ActivateHandler implements StatusChangeHandlerProvider.StatusChangeHandler {
 
+    // Step used which should be performed only once per integration
+    static final String STEP_GITHUB = "github-setup";
```
rhuss: Why is package level visibility bad per se ?

chirino: I've wondered that too.  Guess protected is the closest option.

controllers/src/main/java/com/redhat/ipaas/controllers/integration/StatusChangeHandlerProvider.java
jimmidyson: We use Optionals to identify optional fields in our model. I don't see a problem with that.

PR #366, opened 2017-04-28
ci(circleci): Publish docker images for any branch with summit-demo prefix

jimmidyson:


.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ jobs:
 
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
               DOCKER_IMAGE_TAG="latest"
-            elif [ "${CIRCLE_BRANCH}" == "summit-demo" ]; then
+            elif [[ "${CIRCLE_BRANCH}" =~ "^summit-demo" ]]; then
```
rhuss: lgtm. what is the reason to not require an exact match ? will there be multiple images ?

jimmidyson: Sorry should have said. Yes we will have another branch for Hiram and maybe others (although I don't think so right now). Just because I don't want to update it over weekend, etc if Hiram needs more branches - just cutting a branch with that prefix will be enough.

PR #375, opened 2017-05-25
chore: Last of the rebranding effort

zregvart:
The Versions template file is placed in the com/redhat/ipaas/rest
directory, this moves it into io/syndesis/rest.


jimmidyson:
Thanks!

PR #376, opened 2017-05-25
refactor: alternate Versions strategy

zregvart:
Instead of templating/resource filtering this changes the Version to
return package implementation version when packaged or fixed string
`DEVELOPMENT` when not.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -437,6 +437,17 @@
     <pluginManagement>
       <plugins>
         <plugin>
+          <artifactId>maven-jar-plugin</artifactId>
+          <version>${maven-jar-plugin.version}</version>
+          <configuration>
+            <archive>
+              <manifest>
+                <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+              </manifest>
+            </archive>
+          </configuration>
+        </plugin>
+        <plugin>
           <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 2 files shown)


jimmidyson:
Great idea!

Do you also need to configure the maven-jar-plugin to properly add the manifest entries? Think you can do that like:

```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>
      <manifest>
        <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
      </manifest>
    </archive>
  </configuration>
</plugin>
```

Otherwise there will never be a package version afaik.

zregvart:
retest this please

pom.xml
```diff
@@ -436,6 +436,16 @@
   <build>
     <pluginManagement>
       <plugins>
+      <plugin>
```
jimmidyson: indentation!

pom.xml
```diff
@@ -436,6 +436,16 @@
...
+        <artifactId>maven-jar-plugin</artifactId>
```
jimmidyson: should we be explicit with a version?

jimmidyson:
Thanks! Manually merging until system tests are fixed...

PR #381, opened 2017-05-26
test: Add unit tests for the KeyGenerator

### LINKED ISSUE #380: Add some more unit tests for the KeyGenerator

chirino:
Fixes #380

### DIFF
--- a/core/src/main/java/io/syndesis/core/KeyGenerator.java
+++ b/core/src/main/java/io/syndesis/core/KeyGenerator.java
@@ -25,6 +25,10 @@ import java.util.Random;
  * Generates lexically sortable unique keys based on:
  *
  * https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html
+ *
+ * You can also consider the generated kys to be like UUIDS except:
+ * (1) strictly increment from the generating node's point of view
+ * (2) loosely increment based on relative machine time when viewed across nodes.
  */
 public class KeyGenerator {
 
--- a/core/src/test/java/io/syndesis/core/KeyGeneratorTest.java
+++ b/core/src/test/java/io/syndesis/core/KeyGeneratorTest.java
@@ -20,8 +20,15 @@ import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 
+/**
+ * Unit tests for KeyGenerator
+ */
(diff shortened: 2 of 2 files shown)


chirino:
Hi Kurt, the javadoc has reference to https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html which contains all the detailed information on why it's handy.

This algorithm is what Google is using in Firebase, a massively scaleable JSON based document store similar to jsondb.  Besides being shorter than a UUID, the keys when lexicographically sorted are ordered based on when they were created.  Neat eh?

So if you sort by key, you basically get creation time based sorting.

KurtStam:
Hi @chirino, Links go dead. It'd be good a reference but I'd still like to see the javadoc of the relevant documentation. 

Why do we care about it being order by creation time? I not trying to be difficult but I think it would make sense to keep the codebase as small as possible so if this doesn't solve a specific problem then I'd rather not have it no matter how cool it is.

Also you skipped over my comment of the test itself.

Cheers --Kurt

chirino:
Re the tests: since these keys have an embedded timestamp, these tests try to verify that 2 keys generated in the same ms are not dups.

chirino:
When your adding rows to something like jsondb, with a generated key it's  nice when those added items are returned in order created.  It also lets the implementation work just like firebase does.

KurtStam:
> Re the tests: since these keys have an embedded timestamp, these tests try to verify that 2 keys generated in the same ms are not dups.
ah ok, that wasn't clear to me at all. May want to add a comment in test to this effect. Wouldn't you want to execute the test with some more threads then?

> We use these keys to implement the push() feature in jsondb which is also implemented in firebase. Basically you can 'push' entities into an array. and just like javascript, the entity will be appended to the array. To do that the keys need to always be incrementing.

This would be great as part of the javadoc :).

jimmidyson:
@chirino Care to address @KurtStam's comments?

KurtStam:
LGTM, codacy seems to still be complaining. Cheers --Kurt

KurtStam:
@chirino I'd like to merge this but I think we'll have to wait till the build is working again.

KurtStam:
retest please

KurtStam:
retest this please

PR #383, opened 2017-05-31
Credentials support

zregvart:
This is a work in progress to support credentials, starting with the
easiest parts first :)

See: syndesisio/syndesis-ui#556

Problem:
There is a need for managing credentials (OAuth tokens, username and
passwords, certificates, private keys, ...) that are needed for a
connector to authenticate against a 3rd party resource (service,
database, queue, ...).

1/
First issue is how to store and provide connectors with appropriate
credentials.

2/
Second issue is how to obtain credentials as they can be issued by the
3rd party or require input from the user of the system.

Solution:
1/
Provide API to store and retrieve credentials that solves first issue.
This API can be queried to retrieve manage all credentials or to manage
only credentials applicable to single connector instance.
Selection of applicable credential is done via connector group name
i.e. Camel label, e.g. 'twitter'.

Proposed REST API:
`credentials/{id}` - manages all credentials in the system
`connectors/{id}/credentials/{id}` - manages credentials applicable to
  single connector

2/
Provide API for information on obtaining credentials that solves part of
the second issue. This API will provide information to the UI on how to
obtain credentials (e.g. type: username/password input or link,
properties: list of properties for manual input).

Proposed REST API:
`connectors/{id}/credentials/acquisition` - provides the info on
  obtaining the credential

3/
Provide (Java) plug-in API performing credential acquisition and
endpoint for each plugin to operate in (if needed).

Proposed REST API:
`connectors/{id}/credentials/acquisition/{type}/*` - endpoint for each
  method of credential acquisition

Proposed (Java) plug-in API:
interface CredentialAcquisitor {
  AcquisitionInfo info();
  AcquisitionHandler handler(CredentialRepository);
}

Issues:
Storage of credentials -- k8s secrets?
ACL enforcement -- tie in with Keycloak?

### DIFF
--- a/credential/pom.xml
+++ b/credential/pom.xml
@@ -1,0 +1,191 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (C) 2016 Red Hat, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.syndesis</groupId>
+    <artifactId>syndesis-rest-parent</artifactId>
+    <version>0.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>credential</artifactId>
+  <name>Syndesis REST :: Credential</name>
+
+  <dependencies>
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>model</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>dao</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-beans</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-autoconfigure</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.social</groupId>
+      <artifactId>spring-social-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.social</groupId>
+      <artifactId>spring-social-config</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.immutables</groupId>
+      <artifactId>value</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>findbugs-annotations</artifactId>
+    </dependency>
+
+
+    <dependency>
+      <groupId>javax.validation</groupId>
+      <artifactId>validation-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hibernate.javax.persistence</groupId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
+    </dependency>
+
+    <!-- === Supported services ============================================================== -->
+
+    <dependency>
+      <groupId>com.github.mikegirard</groupId>
+      <artifactId>spring-social-salesforce</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.social</groupId>
+      <artifactId>spring-social-twitter</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <!-- === Test ============================================================================ -->
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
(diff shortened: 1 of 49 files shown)


KurtStam:
@zregvart nice description, can you open an issue for it and add that there too?

zregvart:
> @zregvart nice description, can you open an issue for it and add that there too?

was lazy and just copy-pasted this to #386, good enough to start discussion IMHO.

rhuss:
@zregvart @KurtStam let's merge the design document over. I'm going to review it quickly in a second so it should be available on master, too.

model/src/main/java/io/syndesis/model/Kind.java
```diff
@@ -75,16 +75,16 @@ public String toString() {
...
         }
         return kind;
     }
 
-    public static Kind from(String x) {
-        Kind kind = nameMap.get(x);
+    public static Kind from(final String x) {
+        final Kind kind = nameMap.get(x);
         if( kind == null ) {
             throw new IllegalArgumentException("No matching Lind found.");
```
jimmidyson: Unrelated, but typo: `Lind`

zregvart:
@jimmidyson I think this would be a good time to take a look, I'm going to improve a bit by writing javadocs and adding some unit tests, but overall the design and somewhat the implementation should not change. Thanks 👍

credential/src/main/java/io/syndesis/credential/Credentials.java
```diff
@@ -0,0 +1,251 @@
...
+
+        final ConnectionFactory<?> connectionFactory = factoryFor(providerId);
+
+        final String redirectUrl;
+        final String stateKey;
+        final CredentialFlowState.Builder flowState = new CredentialFlowState.Builder().returnUrl(returnUrl)
+            .connectionId(connectionId).providerId(providerId);
+
+        if (connectionFactory instanceof OAuth1ConnectionFactory) {
+            final OAuth1ConnectionFactory<?> oauth1 = (OAuth1ConnectionFactory<?>) connectionFactory;
```
jimmidyson: How about separating the conditional functionality out into separate methods to make it more readable?

credential/src/main/java/io/syndesis/credential/CredentialProviderRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public <A, T> void addCredentialProvider(final CredentialProvider<A, T> credentialProvider) {
+        final ConnectionFactory<A> connectionFactory = credentialProvider.connectionFactory();
+        final String providerId = connectionFactory.getProviderId();
+
+        providers.put(providerId, credentialProvider);
+    }
+
+    @Override
+    public Applicator<?> getAplicator(final String providerId) {
```
jimmidyson: nit: typo `getAplicator` -> `getApplicator`

credential/src/main/java/io/syndesis/credential/salesforce/SalesforceConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+        public SalesforceApplicator(final SocialProperties socialProperties) {
+            super(socialProperties);
+
+            setClientIdProperty("clientId");
+            setClientSecretProperty("clientSecret");
+            setRefreshTokenProperty("refreshToken");
+        }
+
+        @Override
+        protected Connection.Builder additionalApplication(final Connection.Builder mutableConnection,
```
jimmidyson: Again, I'd prefer to use immutable parameters and return new.

pom.xml
```diff
@@ -186,13 +196,6 @@
         <type>pom</type>
       </dependency>
       <dependency>
-        <groupId>org.infinispan</groupId>
```
jimmidyson: Why remove this BOM and use explicit deps below? Something wrong with the BOM?

rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorCredentialHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Api(value = "credentials")
+public class ConnectorCredentialHandler {
+
+    private final String connectorId;
+
+    private final Credentials credentials;
+
+    public ConnectorCredentialHandler(final Credentials credentials, final String connectorId) {
+        this.credentials = Objects.requireNonNull(credentials, "credentials");
```
jimmidyson: Do we really want to throw NPEs here? How are these going to be represented to the user? This is the same for all `Objects.requireNonNull` calls I guess.

jimmidyson: Wonder if we can use type annotations like `@NonNull` to help at compile time here?

runtime/pom.xml
```diff
@@ -110,8 +110,10 @@
               <title>Syndesis API</title>
               <version>v1</version>
               <description>
-                The Syndesis REST API connects to back-end services on the Syndesis and provides a single entry
-                point for the Syndesis UI. For console developement it can run in off-line mode where it only serves
+                The Syndesis REST API connects to back-end services on the Syndesis and provides a single
+                entry
```
jimmidyson: nit: annoying formatting change :(

zregvart:
> One biggie outstanding for me from this though is how the UI knows what connection fields will be completed by the credential acquisition and hence not hide them on the configure connection page?

The UI should issue a `GET` to `/api/v1/connectors/{connector}/credentials` that will return either [some information](https://github.com/zregvart/syndesis-rest/blob/1956916765dc215917b97907861917d2f052511e/credential/src/main/java/io/syndesis/credential/Credentials.java#L205-L206) (currently not very nice) or [no information](https://github.com/zregvart/syndesis-rest/blob/1956916765dc215917b97907861917d2f052511e/credential/src/main/java/io/syndesis/credential/AcquisitionMethod.java#L30).

credential/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>findbugs-annotations</artifactId>
```
zregvart: The generated code from Immutables contains `@SuppressFBWarnings` annotation, so it's compile/runtime dependency.

credential/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+      <groupId>org.hibernate.javax.persistence</groupId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
+    </dependency>
+
+    <!-- === Supported services ============================================================== -->
+
+    <dependency>
+      <groupId>com.github.mikegirard</groupId>
+      <artifactId>spring-social-salesforce</artifactId>
+      <optional>true</optional>
```
zregvart: The idea would be that `runtime` can choose what providers to bundle and then what providers to configure, so _optional_ gives that power of choice.

credential/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-test</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <scope>test</scope>
```
zregvart: It's used in tests, when you try to touch one of the classes generated by Immutables. At runtime one could (doubtfully) choose another JSR 303 implementation. Seems wrong to pass this choice to clients of _credentials_ and it's not strictly _optional_, so _test_ is something in between. So I'm not sure that _optional_ would be any better than _test_. Not married to it, can change if you think it would be better.

pom.xml
zregvart: The Spring BOM was declared before Infinispan BOM and that caused some Infinispan artifacts to be of that version, when I moved Infinispan BOM before Spring BOM the opposite occured (some of the Spring dependency versions were taken from Infinispan BOM), could not find a more elegant and correct way of making it work 😞

rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorCredentialHandler.java
zregvart: Hmm, I like defensive coding, although this seems to be contentious issue. The Immutables already generate code with `Objects.requireNonNull`, but I can add `@NonNull` annotations instead. Not sure what's our code style on that.

credential/pom.xml
jimmidyson: No it's fine as is.

rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorCredentialHandler.java
jimmidyson: We could have discussed it before you updated! @rhuss thoughts?

pom.xml
jimmidyson: What a mess dependencies always are...

zregvart:
I still want to add more tests and scatter some javadoc over this, and refactor oauth1/2 into separate methods for clarity. Once I'm done with that I'll squash and merge manually if the system tests fail at that point.

jimmidyson:
OK thanks @zregvart!

zregvart:
I need to fix that `ClassNotFoundException` at runtime.

jimmidyson:
Rebase needed... Is this nearly ready to be squashed and merged now?

zregvart:
@jimmidyson rebased, squashed should be OK to merge, let's see the checks pass once again to be sure

PR #390, opened 2017-06-07
Increase timeout for single integration tests to 150s

zregvart:
The basepom default of 30s for a single integration test timeout in basepom, i.e. the tests run using failsafe plugin, is much too small for the integration tests to complete.
Previously in c91ed3454f78f9516eb98613dc21718098497e7d property `basepom.test.timeout` was set, but property `basepom.failsafe.timeout` should have been set.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,8 @@
     <maven.compiler.target>1.8</maven.compiler.target>
     <basepom.test.timeout>300</basepom.test.timeout>
     <basepom.check.skip-all>true</basepom.check.skip-all>
+    <basepom.test.timeout>150</basepom.test.timeout>
+    <basepom.failsafe.timeout>150</basepom.failsafe.timeout>
 
     <syndesis-connectors.version>0.4.4</syndesis-connectors.version>
 
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -233,7 +233,6 @@
                 <!-- workaround suggested in https://github.com/spring-projects/spring-boot/issues/6254 -->
                 <additionalClasspathElement>${project.build.outputDirectory}</additionalClasspathElement>
               </additionalClasspathElements>
-              <forkedProcessTimeoutInSeconds>150</forkedProcessTimeoutInSeconds>
               <systemProperties>
                 <keycloak.http.port>${keycloak.http.port}</keycloak.http.port>
               </systemProperties>


jimmidyson:
Still seeing:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.20:test (default-test) on project project-generator: There was a timeout or other error in the fork -> [Help 1]
```

zregvart:
> Still seeing:
> 
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.20:test (default-test) on project project-generator: There was a timeout or other error in the fork -> [Help 1]

apparently `basepom.test.timeout` is still needed for unit tests run by surefire, set that to 150s also, hopefully should be ok now. Perhaps we should mark some of the unit tests as integration tests?

rhuss:
For me it works now. @zregvart btw, is there a documentation of all properties which influence the base pom behaviour ?

btw, I have the impression that our builds now take considerably longer. 8 minutes is quite a mouthful. Could we separate unit and integration tests ? (e.g. via a profile)

zregvart:
> For me it works now. @zregvart btw, is there a documentation of all properties which influence the base pom behaviour ?

thank you for testing 👍; sorry not much documentation at the moment, I read the [foundation POM](https://github.com/basepom/basepom/blob/master/foundation/pom.xml) to see all the properties, there are some comments inline that help

> btw, I have the impression that our builds now take considerably longer. 8 minutes is quite a mouthful. Could we separate unit and integration tests ? (e.g. via a profile)

Yeah it's doing more so it takes longer, do you see the increase in tests or in the build overall? The build overall is because of all the dependency/static code checks that are now run, the tests can be slower if they were run within the same JVM and because of that caching helped. We could consider running integration tests in parallel (i.e. setting `basepom.failsafe.fork-count` to `1.5C` or similar).

Ultimately as this is a greenfield project we might consider using [Takari smart builder](http://takari.io/book/30-team-maven.html#takari-smart-builder) extension, or running Maven module build concurrently (i.e. `-T 1.5C`) to decrease the build time.

Perhaps as a first step, define a performance budget, for instance not longer that 5 or 10min for the full build, and see what can be shed if we overstep it.

zregvart:
@KurtStam can you take a look at this and approve, from @rhuss feedback it works OK for him as well.

rhuss:
[merge]

zregvart:
[merge]

rhuss:
[merge]

PR #392, opened 2017-06-07
Introduce Maven parallel build to decrease build time

zregvart:
This changes the Maven wrapper configuration to build in parallel,
thread count is set to 1.5C -- the number of running tasks in parallel
is equal to number of cores a machine possesses increased by 50%.
Also tunes the JVM running the Maven by increasing the default maximum
heap size to 512MB and enabling G1 garbage collector for string
de-duplication that should reduce the memory used by Maven due to a
large usage of duplicate strings (repeating GAV strings).

With this build time on my machine was reduced from 05:03 min to 04:17
min (~18% improvement), but I would not expect a radical improvement in
build times due to a (relatively) small number of Maven modules that can
be built in parallel.

If this change is merged it would make troubleshooting a more
complicated as the standard output now includes intertwined output of
several modules being built (from what I can see maximum of 2 at the
moment). This can be helped by configuring logging not to log to
standard output (Keycloak, Spring Boot).

I experimented with running integration tests (run by failsafe) in
parallel, but they fail due to not being able to load Sqlite
(java.lang.UnsatisfiedLinkError: org.sqlite.core.NativeDB._open_utf8([BI)V)
so probably due to the fact that native library cannot be loaded twice
in the same JVM. I also would suspect that the Infinispan caching would
interfere with tests run in parallel causing the tests to become flaky.

Running with Takari smart lifecycle did not improve the build times
dramatically -- they remained more or less the same at 4:26min, I guess
it makes sense using it with larger projects.

One thing to consider would be to include the Takari Concurrent Local
Repository[1] for safe usage of local Maven repositry in concurrent
builds when run with `install` goal.

[1] https://github.com/takari/takari-local-repository


zregvart:
@KurtStam, @jimmidyson, @rhuss any thoughts on this?

rhuss:
@zregvart Does this change affect only the included mvnw, right ? If (as I do) use my locally installed maven I would have to add the options manually.

rhuss:
lgtm, we could try it and when it breaks, the rollback would be trivial.

zregvart:
> @zregvart Does this change affect only the included mvnw, right ? If (as I do) use my locally installed maven I would have to add the options manually.

from [Maven 3.3.1](https://maven.apache.org/docs/3.3.1/release-notes.html#JVM_and_Command_Line_Options) this affects global Maven as well

Issue #384, opened 2017-05-31
Create Camel Component for Twitter (keyword) Search


KurtStam:
Part of https://issues.jboss.org/browse/IPAAS-321 to add support for Twitter Search:

1. Add Camel Component
2. Update the Twitter Search MetaData for TwitterSearch

KurtStam:
@jimmidyson @davsclaus the twitter connector is using the same field for keyword search as for 'complex search'. 

see https://github.com/KurtStam/camel/blob/master/connectors/examples/twitter-search-connector/src/main/resources/camel-connector-schema.json#L27 which is a problem if we want to build a widget as the widget will think ‘string’ and we really want ‘array’. Do we need two connectors if we want to do that, or should we spilt this up into two fields (keywords: array, complex search: string)?

davsclaus:
Just separate the words with comma, eg to search for dog or cat

     keywords=dog,cat

KurtStam:
@davsclaus Yes, but remember the UI builds a widget based on the data type. So for keywords we'd like to see a widget that adds keywords.

davsclaus:
Why not let the citizen user just use a comma. The twitter client on their mobile phone don't have a widget with +/- to add/remove rows to make searching tweets with multiple keywords, they just use comma today.
   
    https://twitter.com/search?q=dog%2Ccat&src=typd
![screen shot 2017-06-05 at 16 40 49](https://cloud.githubusercontent.com/assets/477100/26788525/c305a592-4a0d-11e7-8b30-082e9e5cdfcc.png)

davsclaus:
Anyway from Camel pov, it does not matter how you build the UI, just that the information you feed into Camel is in one option, `keywords=dog,cat`

KurtStam:
@davsclaus can the twitter connector transform array of strings a comma separated string before handing it to the camel component?

davsclaus:
No that is not possible.

As you can see from that camel-connector-schema file each of those row has one value only, not a dynamic array of values.

KurtStam:
@davsclaus ok so the only way yo address this then is to create a second field that takes an array in the camel component, and create a second connector that references this field in the camel-connector-schema.

davsclaus:
No these options cannot really support an array you can specify. Mind that Camel is most often configured in camel uri's, where you use key=value pairs. An array in that format would then be either as we do by using comma as separator, or allow to duplicate the key=value pairs. 

So you do

    keywords=dog&keyword=cat

[...]

PR #393, opened 2017-06-08
Create Camel Component for Twitter (keyword) Search #384

KurtStam:
Note that the twitter-search-connector is new and still needs to be released from the syndesisio/connectors project

### DIFF
--- a/dao/src/main/resources/io/syndesis/dao/deployment.json
+++ b/dao/src/main/resources/io/syndesis/dao/deployment.json
@@ -82,7 +82,47 @@
           "description": "Search tweets based one or more keywords",
           "id": "io.syndesis:twitter-search-connector:latest",
           "camelConnectorGAV": "io.syndesis:twitter-search-connector:@syndesis-connectors.version@",
-          "camelConnectorPrefix": "twitter-search"
+          "camelConnectorPrefix": "twitter-search",
+          "properties": {
+            "filterOld": {
+              "kind": "parameter",
+              "displayName": "Filter Old",
+              "group": "filter",
+              "label": "consumer,filter",
+              "required": false,
+              "type": "boolean",
+              "javaType": "boolean",
+              "deprecated": false,
+              "secret": false,
+              "defaultValue": true,
+              "description": "Filter out old tweets that have previously been polled. This state is stored in memory only and is based on last tweet id."
+            },
+            "keywords": {
+              "kind": "parameter",
+              "displayName": "Keywords",
+              "group": "filter",
+              "label": "consumer,filter",
+              "required": false,
+              "type": "string",
+              "javaType": "java.lang.String",
+              "deprecated": false,
+              "secret": false,
+              "description": "Multiple search values can be separated with comma"
+            },
+           "delay": {
+              "kind": "parameter",
+              "displayName": "Delay",
+              "group": "scheduler",
+              "label": "consumer,scheduler",
+              "required": false,
+              "type": "integer",
+              "javaType": "long",
+              "deprecated": false,
+              "secret": false,
+              "defaultValue": 5000,
+              "description": "Milliseconds before the next poll."
+            }
+          }
         },
         {
           "name": "Send a Tweet",
--- a/project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
+++ b/project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
@@ -27,6 +27,7 @@ import io.syndesis.model.connection.Connector;
 import io.syndesis.model.integration.Integration;
 import io.syndesis.model.integration.Step;
 
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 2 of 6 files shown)


dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -82,7 +82,12 @@
           "description": "Search tweets based one or more keywords",
           "id": "io.syndesis:twitter-search-connector:latest",
           "camelConnectorGAV": "io.syndesis:twitter-search-connector:@syndesis-connectors.version@",
-          "camelConnectorPrefix": "twitter-search"
+          "camelConnectorPrefix": "twitter-search",
+          "properties": {
+          	    "filterOld": { "kind": "parameter", "displayName": "Filter Old", "group": "filter", "label": "consumer,filter", "required": false, "type": "boolean", "javaType": "boolean", "deprecated": false, "secret": false, "defaultValue": true, "description": "Filter out old tweets that has previously been polled. This state is stored in memory only and based on last tweet id." },
```
gashcrumb: nit: `that has` should be `that have`, also should it be `is based on`?

KurtStam: @gashcrumb thx, fixed

rhuss:
What is the status of this PR and what needs to be done in connectors to finally release it ? We should get this functionality merged until the end of this week.

KurtStam:
As stated above it requires a release of the synthesis/connector project.

KurtStam:
This is good to go. Can someone please give it a review/approval?

dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -1,1897 +1,2625 @@
 [
-  {
```
rhuss: It looks like that the formatting has been changed making the diff quite meaningless (and hence not possible to review here). Maybe a tab issue ?

project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
```diff
@@ -48,6 +48,7 @@
             System.setProperty("ivy.message.logger.level", "3");
     }
 
+    private static final String CONNECTOR_VERSION="0.4.5";
```
rhuss: completely agreed that it makes totally sense to extract this as constant. However it would be awesome if this property could be read from property which has been filter-copied by Maven with replacing the version with the current connector version used in the pom.xml. Otherwise its 100% guaranteed that this value will not be updated when connectors get updated.

project-generator/src/main/resources/io/syndesis/project/converter/templates/pom.xml.mustache
```diff
@@ -16,6 +16,7 @@
     <camel.version>@camel.version@</camel.version>
     <funktion.version>@funktion.version@</funktion.version>
     <atlasmap.version>@atlasmap.version@</atlasmap.version>
+    <syndesis-connectors.version>@syndesis-connectors.version@</syndesis-connectors.version>
```
rhuss: Where is this version used ?

dao/src/main/resources/io/syndesis/dao/deployment.json
KurtStam: Good catch, my editor sucks :( and reformatted the entire file even though I only highlighted the added section. Should be fixed now.

project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
KurtStam: Good point - I'm reading it from a (filtered) properties file now.

project-generator/src/main/resources/io/syndesis/project/converter/templates/pom.xml.mustache
KurtStam: It's not AFAI can tell, but these is this test:

https://github.com/KurtStam/syndesis-rest/blob/master/project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java#L150

Not sure why that test is there though, but I thought to play is safe.

rhuss: The test checks for the existance of the pom.xml, not its content. So I'm not sure why the syndesis-connectors.version property within the pom.xml should affect the test. 

The dependency on the connectors must be introduced somewhere else (not via that template). 

If it is not used, could you please remove the property ? if required we can still add it then.

rhuss: ah sorry, misread the test. Its about the file content. But you could remove the property from the 'test-pull-push-pom.xml' fixture, too ?

KurtStam: @rhuss It's actually used in the 'test-pull-push-pom.xml'.

project-generator/src/test/resources/io/syndesis/project/converter/test-pull-push-pom.xml
```diff
@@ -16,6 +16,7 @@
     <camel.version>@camel.version@</camel.version>
     <funktion.version>@funktion.version@</funktion.version>
     <atlasmap.version>@atlasmap.version@</atlasmap.version>
+    <syndesis-connectors.version>@syndesis-connectors.version@</syndesis-connectors.version>
```
rhuss: If you remove this line and remove it from the mustache template, the test should still work, right ? (both lines are added by this PR) because the generator adds the version literally (somehow, don't see it in the template though).

I ask because if you look into the generate pom.xml for an integration, the version is still inserted literally for the connectors and you have a property with no use.

KurtStam: Well that test would work, but the parameter is used in the pom, so then that would be undefined. Maybe I'm missing something.

rhuss: afais the ${syndesis-connectors.version} parameter below is replaced when the resources are copied over with the value from the pom.xml so that when the tests runs its already resolved and no property statement is included but the literal version numbers.

But I could be wrong here, so would be worth a try.

project-generator/src/main/resources/io/syndesis/project/converter/templates/pom.xml.mustache
jimmidyson: Remember that connectors could be at different versions, hence the need to specify the version in the dependency in the POM. I don't think taking this into a single property is safe to do here. If you do want to use properties, then each connector dependency should have it's own separate property for version.

jimmidyson: I've just submitted #414 which properly uses pom properties for other dependencies, but I still think we should directly add the version to the connectors dependencies declarations.

rhuss:
@KurtStam could you please rebase ? I'd like to finish up this story.

KurtStam:
@rhuss Rebase done.

zregvart:
retest this please

Issue #400, opened 2017-06-22
Slow builds


rhuss:
Because of including `basepom` with #377
the build currently still takes far too long (even with the optimisations from #392) for a developer machine. A `mvn clean install` takes 12:00 min here (12:33 min without that optimisation) on a 16 GB 2-year old Mac book running full blast on all 4 cores and with a load of 10 ... 22 (!!) which makes the complete make unusable during the build for doing other things (its even hard to write the text here)
It was used to be less than four minutes.

Also, I get a lot of warnings about parallel builds now.

I don't mind much if CI would take a bit longer, but the current situation will for sure lead to people switching of testing. This is not really usable, especially when we want to attract a community.

All in all, I get the feeling that `basepom` gets overzealous at the expense of developer experience. 

Actually, all those checks (which might make sense) must not be the default. Let a CI build break. But if a build takes longer than 4 minutes (at max !) it's definitely worse than a clean code base.

There is something completely messed up which has to be fixed asap (maybe there are things running twice ?). Can we just remove the base pom or what needs to be done to get our build speedy again ?

// @zregvart @KurtStam @jimmidyson 

(to be fair, normally I would deep dive now to find the issue. would love this, but because of other constraints I can not afford this currently and have to shout out a 'please fix this!' for now ;(

zregvart:
So how about this, keep the basepom, but deactivate most of the plugins, and create a profile that activates them for running in CI?

rhuss:
I'm open for any suggestion and absolutely fine with using different profiles, with the default profile to be fast. Because that's the one which is hit first by any potential developer.

@zregvart Could you do this ? would be awesome !

zregvart:
@rhuss on my machine with #401:

    [INFO] BUILD SUCCESS
    [INFO] ------------------------------------------------------------------------
    [INFO] Total time: 02:35 min
    [INFO] Finished at: 2017-06-22T20:19:31+02:00
    [INFO] Final Memory: 171M/3072M
    [INFO] ------------------------------------------------------------------------

also, you might want to get a new machine 😃

PR #401, opened 2017-06-22
fix: Slow builds (#400)

zregvart:
This disables all checks brought by basepom and adds a Maven profile: `checks` to enable them. Also moves the `ci` profile from `runtime` module to `syndesis-rest-parent` for better visibility and to enable the checks when run in CI.


rhuss:
thanks a lot !

[merge]

PR #403, opened 2017-06-27
Remove pinning of image as the issue has been fixed in the base image

rhuss:


iocanel:
Initially request that we also cleanup the ci profile, but this can happen at a latter point, so changed review to `approved`.

rhuss:
still merging manually as I dont think that the test failure are due to this particular change.

PR #407, opened 2017-06-29
Add missing env variables to maven container

iocanel:


Jenkinsfile
```diff
@@ -1,24 +1,23 @@
 def mavenVersion='3.3.9'
+node {
```
rhuss: Why is this added ?

Jenkinsfile
```diff
@@ -1,24 +1,23 @@
...
+    slave {
+        withOpenshift {
+                withMaven(mavenImage: "maven:${mavenVersion}", 
+		envVars: [ containerEnvVar(key:'GITHUB_OAUTH_CLIENT_ID', value: "${env.GITHUB_OAUTH_CLIENT_ID}"), containerEnvVar(key:'GITHUB_OAUTH_CLIENT_SECRET', value: "${env.GITHUB_OAUTH_CLIENT_SECRET}") ],
```
rhuss: The env vars evaluated by whom ?

Jenkinsfile
iocanel: This is added so that we can have access to env.

Jenkinsfile
iocanel: The env vars are evaluated by the jenkins master. The result is passed to maven container so that they can be used to setup the test environment.

rhuss:
Thanks, let's merge

Issue #422, opened 2017-07-13
The KeyGenerator is not thread safe


zregvart:
Current KeyGenerator implementation is not thread safe, demonstrated by the flakiness of `io.syndesis.jsondb.impl.JsonDBTest::testPush` test.

PR #423, opened 2017-07-13
fix(core): The KeyGenerator is not thread safe (#422)

zregvart:
Current KeyGenerator implementation is not thread safe, demonstrated by the flakiness of `io.syndesis.jsondb.impl.JsonDBTest::testPush` test.

### DIFF
--- a/core/src/main/java/io/syndesis/core/KeyGenerator.java
+++ b/core/src/main/java/io/syndesis/core/KeyGenerator.java
@@ -15,11 +15,15 @@
  */
 package io.syndesis.core;
 
-import org.keycloak.common.util.Base64;
-
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static java.lang.System.currentTimeMillis;
+
+import org.keycloak.common.util.Base64;
 
 /**
  * Generates lexically sortable unique keys based on:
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/syndesis/core/KeyGenerator.java
```diff
@@ -32,40 +36,43 @@
...
+    private static AtomicLong lastTimestamp = new AtomicLong(currentTimeMillis());
     private static final byte randomnessByte;
-    private static long randomnessLong;
-
-    private KeyGenerator() {}
+    private static AtomicLong randomnessLong;
 
     static {
-        Random random = new Random();
+        final Random random = ThreadLocalRandom.current();
```
rhuss: Since this is an static block this method is only called once when the class is initialized. So still when multiple threads later call `createKey` they share the same Random object.

Wouldn't it be better to fetch the `Random` each time afresh in createKey() ?

jimmidyson:
I wonder what you think of moving key to a ULID (as described in https://github.com/alizain/ulid but with Java impls like https://github.com/huxi/sulky/tree/master/sulky-ulid - dependency free btw)?

core/src/main/java/io/syndesis/core/KeyGenerator.java
zregvart: Random values are used only on initialization see the [description of the algorithm](https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html):

    A push ID contains 120 bits of information. The first 48 bits are a timestamp, which both reduces the chance of collision and allows consecutively created push IDs to sort chronologically.
    The timestamp is followed by 72 bits of randomness, which ensures that even two people creating push IDs at the exact same millisecond are extremely unlikely to generate identical IDs.
    One caveat to the randomness is that in order to preserve chronological ordering if a client creates multiple push IDs in the same millisecond, we just 'increment' the random bits by one.

zregvart:
> I wonder what you think of moving key to a ULID (as described in https://github.com/alizain/ulid but with Java impls like https://github.com/huxi/sulky/tree/master/sulky-ulid - dependency free btw)?

Perhaps @chirino can chime in here.

jimmidyson:
agggghhh retest this please

zregvart:
> agggghhh retest this please

I think only system tests failed, right? I think I saw some chatter about that.

chirino:
@jimmidyson @zregvart ULID looks good to me too.  It ticks all my requirements.. but the strings ULID generates are a bit longer right?

PR #430, opened 2017-07-17
chore: Refactor resteasy dependencies to move to spring bean registry scanning for JAX-RS applicatio

### LINKED ISSUE #429: Upgrade resteasy spring boot starter
Also migrate to spring bean registry scanning for JAX-RS application discovery as per https://github.com/paypal/resteasy-spring-boot/blob/master/mds/USAGE.md#registering-jax-rs-application-classes

jimmidyson:
Fixes #429

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -246,21 +246,22 @@
             <goals>
               <goal>unpack</goal>
             </goals>
+            <configuration>
+              <skip>${skipITs}</skip>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.keycloak</groupId>
+                  <artifactId>keycloak-server-dist</artifactId>
+                  <version>${keycloak.version}</version>
+                  <type>tar.gz</type>
+                  <outputDirectory>${project.build.directory}/keycloak</outputDirectory>
+                </artifactItem>
+              </artifactItems>
+            </configuration>
           </execution>
         </executions>
         <configuration>
-          <skip>${skipITs}</skip>
-          <artifactItems>
-            <artifactItem>
-              <groupId>org.keycloak</groupId>
-              <artifactId>keycloak-server-dist</artifactId>
-              <version>${keycloak.version}</version>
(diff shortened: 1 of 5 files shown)


runtime/pom.xml
```diff
@@ -246,21 +246,22 @@
...
+                  <outputDirectory>${project.build.directory}/keycloak</outputDirectory>
+                </artifactItem>
+              </artifactItems>
+            </configuration>
           </execution>
         </executions>
         <configuration>
-          <skip>${skipITs}</skip>
-          <artifactItems>
-            <artifactItem>
```
rhuss: Guess this is not directly to the PR but not needed anymore ?

runtime/src/main/resources/application.yml
```diff
@@ -17,12 +17,6 @@
 deployment:
   file: io/syndesis/dao/deployment.json
 
-resteasy:
```
rhuss: Is it right that the main motivation of this PR is to avoid classpath scanning and so save some startup time ?

runtime/pom.xml
jimmidyson: Oh sorry you're right this snuck in and is unrelated to this PR... I moved the config into the unpack goal [above](https://github.com/syndesisio/syndesis-rest/pull/430/files#diff-48c560bbb8f61d82574c2887b65b498fR250) as `skipITs` for example shouldn't affect dependency analysis...

runtime/src/main/resources/application.yml
jimmidyson: Actually this changes from explicit JAX-RS class specified in config which requires no scanning at all, to default behaviour which is to scan spring registry for all JAX-RS application beans. This allows for easier configuration at the expense of a few ms more scanning time (no full classpath scan so fast).

rhuss: Ah, yes. I was red-green (or better +/- blind ;-)

rhuss:
[merge]

Issue #431, opened 2017-07-17
Finishing OAuth credential flow on a wrong URL


zregvart:
After resource provider callback client is redirected to URL that's relative to API base path (`/api/v1`) not UI. It is expected if given for acquisition `{"returlUrl": "/ui"}` the client should be redirected to `/ui` not `/api/v1/ui`.

jimmidyson:
@zregvart Can you make sure you reference closed issues in the PR description with `fixes #number` please? Then they will get auto closed. We could add that to a PR lint check too possibly.

zregvart:
> @zregvart Can you make sure you reference closed issues in the PR description with fixes #number please? Then they will get auto closed. We could add that to a PR lint check too possibly.

are we doing this in the commit message also?

jimmidyson:
Yes as per conventional changelog guidelines.

PR #432, opened 2017-07-17
fix(credentials): Finishing OAuth credential flow on a wrong URL (#431)

zregvart:
When the UI specifies the return URL in credential acquisition it should be taken as an absolute path not relative path to the API base URI.

This changes to compute the absolute URI at the point the UI submits the relative URI to return to on successful OAuth acquisition.

I've also taken the opportunity to declare Mockito as _test_ dependency.

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionCredentialHandler.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionCredentialHandler.java
@@ -15,6 +15,9 @@
  */
 package io.syndesis.rest.v1.handler.connection;
 
+import java.net.URI;
+import java.net.URISyntaxException;
+
 import javax.annotation.Nonnull;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.Consumes;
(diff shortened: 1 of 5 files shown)


rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionCredentialHandler.java
```diff
@@ -52,6 +55,18 @@ public ConnectionCredentialHandler(@Nonnull final Credentials credentials, @Nonn
...
+
+    protected static URI absoluteTo(final HttpServletRequest httpRequest, final AcquisitionRequest request) {
+        final URI current = URI.create(httpRequest.getRequestURL().toString());
+        final URI returnUrl = request.returnUrl();
+
+        try {
+            return new URI(current.getScheme(), null, current.getHost(), current.getPort(), returnUrl.getPath(),
+                returnUrl.getQuery(), returnUrl.getFragment());
+        } catch (final URISyntaxException e) {
+            throw new IllegalArgumentException(e);
```
jimmidyson: More context might be helpful in the wrapper exception please.

jimmidyson:
Thanks @zregvart!

zregvart:
Fixes #431

Issue #424, opened 2017-07-13
Add REST api to oauth client id/secret resources for the setup page


chirino:

chirino:
This should iterate the connectors to determine which require oauth client id/secrets.  We need to make sure that we can configure these properties at that connector level.

rhuss:
Should we need this API to https://github.com/syndesisio/syndesis-project/issues/28 (which btw the proposal document has not yet started) ?

chirino:
syndesisio/syndesis-project#28 and syndesisio/syndesis-project#13 seem like they are dups of each other no?

rhuss:
@chirino The difference is that https://github.com/syndesisio/syndesis-project/issues/13 is the epic (user story), 28 is the creation of the design proposal (as task attached to this epic).

It is following the workflow I described in https://github.com/rhuss/syndesis.io/blob/4d1b5a284cd0e9359fb49f03ce8f48fe9e7b6004/content/community/engineering_guidelines.md#development-workflow which is currently under review.

zregvart:
This is implemented with #442

Issue #373, opened 2017-05-23
REST apis must not send credentials back to the client


chirino:
I think it might be a good idea if the REST apis which set and get sensitive data was structured so you can only set those sensitive properties but never get them back via the REST apis.

chirino:
We also might need to consider encrypting these values while they are at rest in the DB.

gashcrumb:
Not sure that this would currently affect the UI really except with the connection configuration, even then we don't really need to fetch it.  That connection object gets embedded in the integration object, but again the UI doesn't really need to pass along secrets in that case either...

dsimansk:
I wonder what is conceptually different between storing Secrets as OpenShift resource where you can still retrieve them and storing our Connection secret. In my opinion there should be still an user with appropriate role that can retrieve that credential. Maybe those credentials shouldn't be exposed directly on Connection GET query, but nested/hidden(a bit)?

jimmidyson:
This looks like we might want to think about having different types for properties: something like `Value` and `SecretValue`, where `SecretValue` is a reference to a secret (another type in DB, possibly encrypted) value rather than stored in the `Connection` object in the DB. Then when we're serializing in the GET `Connection` a `SecretValue` would serialize as a fixed string, like `********` or something.

chirino:
We already know which properties are [sensitive](https://github.com/syndesisio/syndesis-rest/blob/master/model/src/main/java/io/syndesis/model/connection/ConfigurationProperty.java#L33).  Can't we just use that?

jimmidyson:
I feel that if something is sensitive then it should be more explicit when requesting to view it. That way we could start to introduce some more permission levels in future.

I also would prefer the fact that something is sensitive to be stored with the data, rather than just a raw String->String map and having to read back from the relevant ConfigurationProperty if it's meant to be sensitve/encrypted.

chirino:
@jimmidyson I think that would have ui impact right @gashcrumb?  The UI cant assume the values are Strings, you might get an object in there for the secret fields.

gashcrumb:
Yeah, most likely.

KurtStam:
@chirino is this still an issue? There have been quite a few commits related to this. Do you know the current status?

chirino:
I've not checked.

rhuss:
Yes, it's still an issue.

chirino:
So this is mostly fixed now, except for the secret configured for the simpledb connector that the DB is initialized with.

PR #433, opened 2017-07-17
Issue #424 Add REST api to oauth client id/secret resources for the setup page

chirino:
Also:
- Adds tags field to properties
- Adds configuredProperties to Connectors
- Increments the schema version

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
@@ -1,0 +1,141 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.rest.v1.handler.setup;
+
+import io.swagger.annotations.Api;
+import io.syndesis.core.SuppressFBWarnings;
+import io.syndesis.dao.manager.DataManager;
+import io.syndesis.model.connection.ConfigurationProperty;
+import io.syndesis.model.connection.Connector;
+import org.springframework.stereotype.Component;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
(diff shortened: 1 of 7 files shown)


rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -0,0 +1,112 @@
...
+
+        List<Connector> items = dataMgr.fetchAll(Connector.class).getItems();
+        items.forEach(connector -> {
+            if (isOauthConnector(connector)) {
+                OAuthApp app = new OAuthApp();
+                app.id = connector.getId().get();
+                app.name = connector.getName();
+                app.icon = connector.getIcon();
+                app.clientId = getPropertyTaggedAs(connector, "oauth-client-id");
+                app.clientSecret = getPropertyTaggedAs(connector, "oauth-client-secret");
```
zregvart: Would this expose secret to the UI? (see  #373)

chirino: yeah. this would need to be addressed as part of #373 too.

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final DataManager dataMgr;
+
+    public OAuthAppHandler(DataManager dataMgr) {
+        this.dataMgr = dataMgr;
+    }
+
+    // Since this a a view model DTO, and not a domain model lets define it here instead
+    // of placing it into the model module.
+    public static final class OAuthApp {
```
jimmidyson: Recommend constructor args and getters here.

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -0,0 +1,141 @@
...
+    public OAuthAppHandler(DataManager dataMgr) {
+        this.dataMgr = dataMgr;
+    }
+
+    // Since this a a view model DTO, and not a domain model lets define it here instead
+    // of placing it into the model module.
+    @SuppressFBWarnings(
+        value = "URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD",
+        justification = "All fields are encode by jackson and sent to the UI")
+    public static final class OAuthApp {
```
zregvart: Would it make sense to use Immutables nonetheless and have a `from(Connector)` method?

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private boolean isOauthConnector(Connector connector) {
+        TreeSet EMPTY = new TreeSet();
+        return connector.getProperties().values().stream().anyMatch(x -> {
+                return x.getTags().orElse(EMPTY).contains("oauth-client-id");
+            }
+        );
+    }
+
+    private String getPropertyTaggedAs(Connector connector, String name) {
```
zregvart: This might be useful on the Connector interface also, I think it can be used in Credentials to further simplify the applicators (can be worked on in another PR)

Issue #435, opened 2017-07-18
`state` parameter is not transfered for OAuth1


zregvart:
When authenticating against an OAuth1 resource provider (Twitter) the `state` parameter needed for accessing credential flow state is not passed to the credentials callback.

rhuss:
I _think_ this is the issue related to #435 which has been merged. Closing it therefor, feel free to reopen.

PR #437, opened 2017-07-18
fix(credentials): `state` parameter is not transfered for OAuth1 (#435)

zregvart:
With this change the `state` parameter is persisted in the HTTP session.

I would like to think about this a bit, submitted for other opinions.

Crux:
 - if we're now using HTTP session for OAuth1 why would we use Infinispan at all
 - any trick on remaining _stateless_ with OAuth1?


rhuss:
retest this please

PR #441, opened 2017-07-19
feat(rest): Parameter validation (#440)

### LINKED ISSUE #440: Parameter validation
Currently we don't have parameter validation implemented, the credentials implementation relies on syntax validation of one parameter (`returnUrl` needs to be a relative URI to prevent open redirect vulnerability). Having at least `@NotNull` on parameters would be more helpful.

zregvart:
This configures RESTEasy to validate parameters in accordance to JSR 303 (Bean Validation). It also adds JSR 303 annotations to interfaces in `io.syndesis.rest.v1.operations` package so that the standard CRUD method parameters and to other *Handler implementations so that they are all validated.

A new ExceptionMapper was added to provide JSON output in case of constraint validation.

With the specific way the exceptions are handled in RESTEasy (handler for [Reader|Writter]Exception is not looked up the inheritance chain) two ExceptionMappers to handle those are added.

And with those changes in the exception handling it made sense to create a base class for ExceptionMappers.

The only way to for the validations to work with Immutables is to use the standard Java Beans method signatures (get/set) as there is currently no way to specify an annotation on the field with Immutables[1].

There is still more work to be done on validations, for instance deeper syntax validations might be beneficial (e.g. disallow certain characters in IDs to prevent injection).

fixes #440

[1] https://github.com/immutables/immutables/issues/475

### DIFF
--- a/credential/src/main/java/io/syndesis/credential/AcquisitionRequest.java
+++ b/credential/src/main/java/io/syndesis/credential/AcquisitionRequest.java
@@ -25,6 +25,7 @@ import org.springframework.validation.annotation.Validated;
 @Value.Immutable
 @JsonDeserialize(builder = AcquisitionRequest.Builder.class)
 @Validated
+@Value.Style(passAnnotations = RelativeUri.class)
 public interface AcquisitionRequest {
 
     class Builder extends ImmutableAcquisitionRequest.Builder {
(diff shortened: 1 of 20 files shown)


zregvart:
Thanks @iocanel 🥇

PR #442, opened 2017-07-19
Integrated the global settings for oauth apps issues #424 and syndesisio/syndesis-ui#556.

chirino:
When a user updates the oauth id/secret for a connection in the settings page, the oauth credentials registry now gets updated.

### DIFF
--- a/credential/pom.xml
+++ b/credential/pom.xml
@@ -121,13 +121,11 @@
     <dependency>
       <groupId>com.github.mikegirard</groupId>
       <artifactId>spring-social-salesforce</artifactId>
-      <optional>true</optional>
     </dependency>
 
     <dependency>
       <groupId>org.springframework.social</groupId>
       <artifactId>spring-social-twitter</artifactId>
-      <optional>true</optional>
     </dependency>
 
     <!-- === Test ============================================================================ -->
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
@@ -15,12 +15,15 @@
  */
 package io.syndesis.rest.v1.handler.setup;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeSet;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiParam;
+import io.syndesis.core.SuppressFBWarnings;
+import io.syndesis.dao.manager.DataManager;
+import io.syndesis.model.connection.ConfigurationProperty;
+import io.syndesis.model.connection.Connector;
+import org.springframework.stereotype.Component;
 
+import javax.persistence.EntityNotFoundException;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.Consumes;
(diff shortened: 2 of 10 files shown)


chirino:
retest this please

runtime/src/test/java/io/syndesis/runtime/TestSupport.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package io.syndesis.runtime;
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+/**
+ * This class defines some methods that can assist you in writing tests
+ */
+public class TestSupport {
```
zregvart: Perhaps we could use [Awaitility](https://github.com/awaitility/awaitility)

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppToCredentialsBridge.java
```diff
@@ -0,0 +1,136 @@
...
+            // Not all connectors will be oauth apps..
+            return;
+        }
+        registerCredentialProvider(oAuthApp);
+    }
+
+    public void registerCredentialProvider(OAuthAppHandler.OAuthApp app) {
+        // is the app configured?
+        if( isSet(app.clientId) && isSet(app.clientSecret) ) {
+            if( "twitter".equals(app.id) ) {
```
zregvart: We could do it this way, or we could bring back the `CredentialProviderConfigurer`s that we chucked out in #439, and have the `id` property there, so they can serve as factories of `CredentialProvider`s, so that we don't have to explicitly check for each supported configuration in the future. Perhaps refine this in another PR.

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -128,7 +150,7 @@ private String getPropertyTaggedAs(Connector connector, String name) {
         return null;
     }
 
-    private void setPropertyTaggedAs(Connector connector, Map<String, String> configuredProperties, String name, String value) {
+    private static void setPropertyTaggedAs(Connector connector, Map<String, String> configuredProperties, String name, String value) {
         TreeSet EMPTY = new TreeSet();
```
zregvart: Just noticed this `EMPTY` from the previous PR, perhaps use a constant from `java.util.Collections` instead (can be refined in a new PR).

zregvart:
@chirino can you use [Conventional Commit messages](https://syndesis.io/community/engineering_guidelines/#commit-messages-should-follow-the-conventional-commits-standardized-commit-message-format) 🖖

runtime/src/test/java/io/syndesis/runtime/TestSupport.java
chirino: That does look nicer... I'll switch

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
chirino: Dont think you can since it has to be a Sorted set.

chirino:
retest this please

credential/pom.xml
```diff
@@ -121,13 +121,11 @@
     <dependency>
       <groupId>com.github.mikegirard</groupId>
       <artifactId>spring-social-salesforce</artifactId>
-      <optional>true</optional>
```
zregvart: We wanted these to be optional, so that the `runtime` module explicitly defines what 3rd party services it supports.

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
zregvart: Sure it does :) [Collections.emptySortedSet()](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptySortedSet--) (for new PR)

rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppToCredentialsBridge.java
zregvart: (for new PR)

rhuss:
@chirino would love to merge the PR. Any chance to give it some love today and adress the review comments ?

Issue #457, opened 2017-07-21
State handling in OAuth


zregvart:
Right now we have non-uniform way of handling state for OAuth, for OAuth1 we use cookies (wonder if that works with Ajax calls) to store `state` key, for OAuth2 the `state` parameter is part of the flow. The flow state is persisted in Infinispan.
What we should definitely do is to consolidate those two ways into a single way of persisting flow state and referencing the `state` (key) parameter. Whole state in MAC-ed cookies seems like a way to go about it.

PR #459, opened 2017-07-25
OAuth client side state

zregvart:
This is part of the work for #457, this can be viewed independently, so I'm creating a PR for early review/merge.


rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
```diff
@@ -0,0 +1,291 @@
...
+    private final Supplier<byte[]> ivSource;
+
+    private final Function<Object, byte[]> serialization;
+
+    private final long timeout;
+
+    private final Supplier<Long> timeSource;
+
+    protected static final class RandomIvSource implements Supplier<byte[]> {
+        private static final ThreadLocal<SecureRandom> RANDOM = new ThreadLocal<SecureRandom>() {
```
jimmidyson: Isn't SecureRandom threadsafe? Do we really want to initialize (expensive) one for every thread we use this on?

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
```diff
@@ -0,0 +1,291 @@
...
+import javax.crypto.spec.IvParameterSpec;
+import javax.ws.rs.core.Cookie;
+import javax.ws.rs.core.NewCookie;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
+
+import org.msgpack.jackson.dataformat.MessagePackFactory;
```
jimmidyson: How much do we save using MessagePack vs JSON?

jimmidyson: i.e. is it worth introducing a new dependency for this?

runtime/src/main/resources/application.yml
```diff
@@ -80,3 +80,11 @@ generator:
 controllers:
   integration:
     enabled: true
+
+client:
+  state:
+    authenticationAlgorithm: HmacSHA1
```
jimmidyson: Did you mean to add these as default settings? How about leaving these out of here?

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
zregvart: Looking at empty OAuth2CredentialFlowState (current working, ex CredentialFlowState) it's 30% difference (80 bytes JSON, 56 bytes MessagePack). We have 2.8KB of space in the Cookie to work with. If it's an issue easily removed 👂

runtime/src/main/resources/application.yml
zregvart: I'm don't fully understand, you mean have them as default so that they don't need to be specified or remove them entirely?

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
zregvart: Yeah it is, this pattern is something I usually do, but now I remembered that we run in Linux containers so it will go to `/dev/urandom`, it would block otherwise on Windows and that makes it needed. Good catch 👍

runtime/src/main/resources/application.yml
jimmidyson: Sorry what I mean was, do we need this configuration in the application.yml that is bundled with the app at build time? Surely this is runtime configuration only and as such should be provided by the environment (via the syndesis-rest configmap in this case)? I don't like having stuff like encryption keys as defaults as it increases the chance of a publicly readable encryption key accidentally being used in a real deployment, e.g. if the configmap doesn't have the right configuration.

jimmidyson: So I mean remove them entirely.

zregvart: Yeah, that's what I was planing on doing, but in a separate PR as this needs some thinking behind it. Right now you can override these via environment variables. Not sure if we can have them dynamically set in the OpenShift templates.

Things to think about:
 - do we need this statically or dynamically (think regenerate keys every n days)
 - do we do this via k8s secrets
 - if dynamically how to handle the transition (need n-1 keys)
 - if dynamically then we need key distribution (spring cloud k8s, but it reloads the app context)

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
jimmidyson: I'm not fussed either way to be honest, msgpack format would bring in an extra 2 dependencies but that's no biggie. So I'll leave it up to you :)

runtime/src/main/resources/application.yml
jimmidyson: I still would like these removed from the committed codebase due to the potential problems I mentioned above.

We could prevent startup if these values aren't set perhaps so we know they have to be set in prod deployments?

Keys should be regenerated, short lived keys are great, but you're right we need to think about how we do that. I'd recommend we do a rolling deployment rather than updating in place, but we also need to figure out how we accept old encrypted cookies and re-encrypt on next request if we do that. But this can be future PR work IMO.

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
zregvart: Yeah, let's use it when we see it's needed.

zregvart:
I'd like to do more manual testing before I declare this done, but it should not change much so reviews are welcome :+1:

rest/src/main/java/io/syndesis/rest/v1/state/ClientSideState.java
rhuss: BTW we have also think about productisation. As I've been told, every single dependency needs to be productised (i.e. needs to be build by red hat). See https://issues.jboss.org/browse/ENTESB-6963 for details and the dependency list we already have.

zregvart:
Manually tested with Salesforce, works 👍

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
```diff
@@ -17,27 +17,91 @@
...
 
     URI getReturnUrl();
 
-    String getKey();
+    default String persistenceKey() {
+        return statePrefix() + getKey();
+    }
+
+    default String statePrefix() {
+        if (this instanceof OAuth1CredentialFlowState) {
```
jimmidyson: How about moving these down into the specific interfaces?

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
```diff
@@ -17,27 +17,91 @@
...
+            return OAUTH1_CREDENTIAL_PREFIX;
+        } else if (this instanceof OAuth2CredentialFlowState) {
+            return OAUTH2_CREDENTIAL_PREFIX;
+        }
+
+        throw new IllegalStateException("Unsupported credential flow state implementation: " + this);
+    }
+
+    default Type type() {
+        if (this instanceof OAuth1CredentialFlowState) {
```
jimmidyson: Same here: How about moving these down into the specific interfaces?

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
```diff
@@ -17,27 +17,91 @@
...
+        }
+
+        throw new IllegalStateException("Unsupported credential flow state implementation: " + this);
+    }
 
-    OAuthToken getToken();
+    CredentialFlowState updateFrom(HttpServletRequest request);
+
+    static Builder builderFor(final ConnectionFactory<?> connectionFactory) {
+        if (connectionFactory instanceof OAuth1ConnectionFactory) {
```
jimmidyson: And here... How about moving these down into the specific interfaces?

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
```diff
@@ -17,27 +17,91 @@
...
+            return new OAuth1CredentialFlowState.Builder();
+        } else if (connectionFactory instanceof OAuth2ConnectionFactory) {
+            return new OAuth2CredentialFlowState.Builder();
+        }
+
+        throw new IllegalStateException("Unsupported connection factory implementation: " + connectionFactory);
+    }
+
+    static Class<? extends CredentialFlowState> typeForName(final String name) {
```
jimmidyson: Why is this necessary? Isn't jackson smart enough to do (de)serialize appropriately for subtypes?

credential/src/main/java/io/syndesis/credential/Credentials.java
```diff
@@ -45,56 +38,43 @@
...
 
         final String redirectUrl;
-        final CredentialFlowState.Builder flowStateBuilder = new CredentialFlowState.Builder().returnUrl(returnUrl)
-            .connectionId(connectionId).providerId(providerId);
+        final CredentialFlowState.Builder flowStateBuilder = CredentialFlowState.builderFor(connectionFactory)
+            .returnUrl(returnUrl).connectionId(connectionId).providerId(providerId);
 
         if (connectionFactory instanceof OAuth1ConnectionFactory) {
-            redirectUrl = prepareOAuth1((OAuth1ConnectionFactory<?>) connectionFactory, flowStateBuilder, request);
+            redirectUrl = prepareOAuth1((OAuth1ConnectionFactory<?>) connectionFactory,
```
jimmidyson: Why not move all these `prepareOAuth{1,2}` and similar methods into relevant connection factories?

pom.xml
```diff
@@ -543,6 +543,18 @@
         <version>3.2.1</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.msgpack</groupId>
```
jimmidyson: So you decided to keep these? That's cool, just misunderstood what you said before, but no worries.

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
zregvart: yeah, silly me, good catch 👍

credential/src/main/java/io/syndesis/credential/CredentialFlowState.java
zregvart: I think I need to give Jackson a concrete type, but I did fiddle with the annotations so it could be leftover from when id did not work for me. I think I need a unit test for that.

credential/src/main/java/io/syndesis/credential/Credentials.java
zregvart: Good point, will try 👍

pom.xml
zregvart: Oh, no, they're out, you know how I like to force push commits 😈

zregvart:
I think its well worth the effort to address the comments from the review in this PR, it should not be a lot of work.

pom.xml
zregvart: Yeah, forgot to remove it from the parent

zregvart:
So with the latest commit there is a definite split between OAuth1 and OAuth2 implementations and this I think improved greatly the simplicity of the implementation.

@jimmidyson could you have another look?

zregvart:
I think the build is failing because deployment descriptor is missing environment variables trying to catch pod failing in system tests to confirm.

See syndesisio/syndesis-openshift-templates#63

zregvart:
Indeed:

    org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'credentialHandler' defined in URL [jar:file:/deployments/runtime.jar!/BOOT-INF/lib/rest-0.1-SNAPSHOT.jar!/io/syndesis/rest/v1/handler/credential/CredentialHandler.class]: Unsatisfied dependency expressed through constructor parameter 1; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'clientSideState' defined in io.syndesis.runtime.Application: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.syndesis.rest.v1.state.ClientSideState]: Factory method 'clientSideState' threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Client side state persistence configuration is not valid, please set
        CLIENT_STATE_AUTHENTICATION_ALGORITHM
        CLIENT_STATE_AUTHENTICATION_KEY
        CLIENT_STATE_ENCRYPTION_ALGORITHM
        CLIENT_STATE_ENCRYPTION_KEY
        CLIENT_STATE_TID
    environment variables.

credential/src/main/java/io/syndesis/credential/OAuth1CredentialFlowState.java
```diff
@@ -0,0 +1,89 @@
...
+    final class Builder extends ImmutableOAuth1CredentialFlowState.Builder implements CredentialFlowState.Builder {
+
+        @Override
+        public CredentialFlowState.Builder withAll(final CredentialFlowState state) {
+            return createFrom(state);
+        }
+
+    }
+
+    final class OAuthTokeConverter implements Converter<Map<String, String>, OAuthToken> {
```
jimmidyson: nit: typo `Toke` -> `Token`

rest/src/main/java/io/syndesis/rest/v1/state/Edition.java
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.rest.v1.state;
+
+/**
+ * One configuration edition of settings used to store state on the client.
+ */
+public abstract class Edition {
```
jimmidyson: Bit confused by the name? Could you explain why it's called `Edition`?

zregvart: Yeah I was trying to find a good name, the idea always was to periodically regenerate the keys, then each new generation would have `tid=tid+1`, so as they roll in time I thought _Edition_ would be a good name for it.

zregvart:
Build succeeded

squashing...

PR #426, opened 2017-07-14
chore(clean): Cleanup code by way of trivial fixes

zregvart:
Trivial to do code cleanup fixes:

 * usage or raw (unparameterized) generic types
 * missing `@Override` annotation
 * _leaky_ resource usage
 * throws declarations from methods that do not throw the specific exception
 * unused fields/parameters/methods
 * trivial spelling mistakes
 * use of `"UTF-8"` or `Charset.forName("UTF-8")` instead of `StandardCharsets.UTF_8`
 * unnecessary casts

PR #461, opened 2017-07-26
fix(persistence): Caching and returning wrong entity from `DataManger::create`

zregvart:
In the code cleanup (#426) result of the `DataManger::create` was wrongly assigned and cached as the given entity, it should be the `entityToCreate` thats being assigned if the `id` is not present.

Fixes syndesisio/syndesis-ui#662

### DIFF
--- a/dao/pom.xml
+++ b/dao/pom.xml
@@ -108,6 +108,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
   </dependencies>
 
   <build>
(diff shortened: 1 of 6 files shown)


zregvart:
retest this please

dao/pom.xml
```diff
@@ -108,6 +108,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
rhuss: <scope>test</scope> is in the dependencyManagement, I assume

zregvart: sure is:

    ...
    [INFO] \- org.assertj:assertj-core:jar:3.6.2:test

Issue #395, opened 2017-06-09
Move from using GitHub API to normal git workflow


rhuss:
At the moment we use the GitHub API, which results in one commit per file (because we are using the Github REST API). 

We need to create the Git repo locally on a volume and then push it out to GitHub. We should the way how we do it with the way how `openshiftio/launchpad-missioncontrol` does it

See https://github.com/openshiftio/launchpad-missioncontrol/tree/master/services/github-service-impl/src/main/java/io/openshift/appdev/missioncontrol/service/github as a starting point.

KurtStam:
@rhuss So if we're no longer going to use the REST api, then what will be using for authentication to GitHub?

rhuss:
That's a good question, and we probably should have a look how the openshift.io launchpad does it. See the link above as a starting point. It probably depends on how the library we want to use handles authentication, but it's probably also via an initial oauth flow. Let me research that a bit (but you might also want to look into the code).

@jimmidyson do you have any idea how this could work?

KurtStam:
I think I would prefer public key authentication off possible so that it works for any git implementation. I'm going to check to see if the GitHub api supports uploading public keys.

rhuss:
No, not yet. And I'm afraid that @kcbabo is not very keen on having the citizen user to deal with public / private key management. 

We could create a keypair on the fly though, do the upload and store the private key in the DB. If require we could let an expert user override this via the UI in the gloabl settings section. But I think this is also a story for UXD.

I think this would be a nice solution.

KurtStam:
@rhuss that's what I was going for, creating one on the fly for the citizen user (if not already present) and overridable if needed.

rhuss:
I think thats a good approach. We can even start without UI by doing only the automatic setup and add this later to the UI.

rhuss:
Is it how the launchpad does it, too ? (haven't had a loook yet)

KurtStam:
@rhuss - the code doesn't seem to deal with authentication, so maybe it assumes it's dealt with elsewhere (i.e. by using keys)

rhuss:
@KurtStam any progress on this story ? Is there something I can help ?

KurtStam:
@rhuss I need to help defining the requirements around this. Here are some questions for you:

[...]

PR #465, opened 2017-07-26
move to normal git flow #395

KurtStam:
Note that the integration test needs an AUTH_TOKEN. This should go in the maven settings.xml. How do we get this on the CI server?


iocanel:
retest this please

jimmidyson:
FYI build failure:

```
[INFO] Checking licenses...
[WARNING] Missing header in: /home/jenkins/workspace/syndesis-rest_PR-465-EDC46WIXSYG6LPPSSUBNJBTUHZUGXQY57VT7PXEWUPMYK4OKC53A/github/src/test/resources/sample-github-project/pom.xml
```

zregvart:
@KurtStam if the resources in `github/src/test/resources` cannot be modified (i.e. they are generated from `project-generator`) perhaps just add `github/src/test/resources/sample-github-project/pom.xml to `excludes` of the license check plugin?

http://code.mycila.com/license-maven-plugin/reports/3.0/check-mojo.html#excludes

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Commits and pushes the files to the Git Repository
+     * 
+     * @param repo- an existing github repository
+     * @param message- commit message
+     * @param files- map of file paths along with their content
+     * @param credentials- Git credentials, for example username/password, authToken, personal access token
+     * @return RevCommit, the commit info
+     */
+    public RevCommit createFiles(String repoHtmlUrl, String repoName, String message, Map<String, byte[]> files,
```
rhuss: Maybe a bit nit-picky but the javadoc params do not correspond the the signature. This is one of the problem with javadoc that it gets stale (even in the initial PR ;-), but happy that you document.

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+            UsernamePasswordCredentialsProvider credentials) {
+
+        try {
+            // create temporary directory
+            Path workingDir = Files.createTempDirectory(repoName);
+            if (log.isDebugEnabled()) log.debug("Created temporary directory {}", workingDir.toString());
```
rhuss: up to now we always used braces for `if`s, even for one liner. Don't want to insist on it, but would be cool if we could stay consistent. There will be very likely a checkstyle rule which will complain, too, I guess.

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+     * @param files- map of file paths along with their content
+     * @param credentials- Git credentials, for example username/password, authToken, personal access token
+     * @return RevCommit, the commit info
+     */
+    public RevCommit createFiles(String repoHtmlUrl, String repoName, String message, Map<String, byte[]> files,
+            UsernamePasswordCredentialsProvider credentials) {
+
+        try {
+            // create temporary directory
+            Path workingDir = Files.createTempDirectory(repoName);
```
rhuss: I wonder whether we should think about a dedicated hostPath volume instead of creating the layer in the container. wonder whether this has any performance implications ? // @jimmidyson

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+    private void writeFiles(Path workingDir, Map<String, byte[]> files) throws IOException {
+        for (String filePath : files.keySet()) {
+            File file = new File(workingDir.toString() + File.separator + filePath);
+            file.getParentFile().mkdirs();
+            Files.write(file.toPath(), files.get(filePath));
+        }
+    }
+        
+    private RevCommit commitAndPush(Git git, String message, UsernamePasswordCredentialsProvider credentials) 
+            throws NoFilepatternException, GitAPIException {
```
rhuss: NoFilePatternException can be removed here

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+     * Write files to the file system
+     * 
+     * @param workingDir
+     * @param files
+     * @throws IOException
+     */
+    private void writeFiles(Path workingDir, Map<String, byte[]> files) throws IOException {
+        for (String filePath : files.keySet()) {
+            File file = new File(workingDir.toString() + File.separator + filePath);
+            file.getParentFile().mkdirs();
```
rhuss: Shouldn't the return value be checked here and an exception thrown if its false ?

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Write files to the file system
+     * 
+     * @param workingDir
+     * @param files
+     * @throws IOException
+     */
+    private void writeFiles(Path workingDir, Map<String, byte[]> files) throws IOException {
+        for (String filePath : files.keySet()) {
+            File file = new File(workingDir.toString() + File.separator + filePath);
```
rhuss: would prefer

```java
new File(workingDir.toFile(),filePath);
```

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,131 @@
...
+        if (log.isDebugEnabled()) log.debug("git add all file");
+
+        // git commit
+        RevCommit commit = git.commit().setMessage(message).call();
+        log.info("git commit id {}", commit.getId());
+
+        // git push -f, not merging but simply forcing the push (for now)
+        Iterable<PushResult> pushResult = git.push().setCredentialsProvider(credentials).setForce(true).call();
+        if (!pushResult.iterator().next().getMessages().equals("")) 
+            log.warn("git push messages: {}", pushResult.iterator().next().getMessages());
```
rhuss: please add braces here. What's the purpose of this warning ?

github/src/main/java/io/syndesis/github/GitHubServiceImpl.java
```diff
@@ -96,27 +94,23 @@ private Repository getRepository(String name) throws IOException {
...
-        for (Map.Entry<String, byte[]> entry : files.entrySet()) {
-            // Wait a bit to let GitHub catch up
-            // See http://stackoverflow.com/questions/19576601/github-api-issue-with-file-upload for details
-            sleep(1000L);
-            createOrUpdateFile(repo, message, entry.getKey(), entry.getValue());
+        GitWorkflow gitWorkflow = new GitWorkflow();
+        Repository repository = getRepository(repo.getName());
+        if (repository == null) {
+            repository = createRepo(repo.getName());
+            gitWorkflow.createFiles(repo.getHtmlUrl(), repo.getName(), message, files, new UsernamePasswordCredentialsProvider(Tokens.getAuthenticationToken(), "") );
```
rhuss: So the KeyCloak authentication token is used as username and an empty string as password. Could you please elaborate a bit how this finally work to add the GitHub OAuth token ?

github/src/main/java/io/syndesis/git/GitWorkflow.java
jimmidyson: Certainly not `hostPath` volume - not supported by OSO I would assume, plus other issues that make it generally unusable outside of dev (like not inheriting ownership from pod security context).

I wonder if we could get away with in-memory for now? We could use an memory-backed emptyDir volume for now and then think about switching to a real volume in future?

rhuss:
@KurtStam any chance that you could address the review feedback before PTO ?

rhuss:
@Kurtstam Hope you dont mind that I'm just fixing the CI failure (which consists of mixture of license-check, fundbugs and checkstyle errors).

It is a good idea to always use profile '-Pci' before creating the PR as this is the profile used by CI, too.

rhuss:
retest

rhuss:
`-Pchecks` is also fine (as @zregvart just told me :)

rhuss:
@KurtStam test's fixed. don't forget to pull again from the branch to get my changes.

github/src/main/java/io/syndesis/git/GitWorkflow.java
KurtStam: @rhuss I guess I fail to see what the issue is using a tempDir, most apps write things to the tmp dir. We only need it there a very small amount of time. @jimmidyson a in-memory file system would work, but I'd rather keep it simple.

rhuss: Not really an issue. just thought about performance implications because of the layered container file system (which might have other issues) but agree that we should keep simple.

just asking ;-)

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
```diff
@@ -0,0 +1,151 @@
...
+    private static GitHubClient client = null;
+    private static GitHubService githubService = null;
+
+    @BeforeClass
+    public static void before() throws IOException {
+
+        // TODO property should be defined in your maven settings.xml
+        authToken = System.getProperties().getProperty("github-oauth-token");
+
+        Assert.assertNotNull(authToken);
```
zregvart: Perhaps we could use assume here

KurtStam:
I'm guessing this has to do with the settings.xml on Jenkins. (see the github/README.md). However I only add the profile section. I'm not quite sure why the nexus url was working before and now not anymore. @zregvart @jimmidyson do you know?

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
```diff
@@ -62,13 +68,44 @@ public void before() throws IOException {
...
+        webserver.start(1234);
+
+        webserver.expect().get().withPath("/auth/realms/syndesis-it/broker/github/token").andReturn(
+            200,
+            "{\"access_token\":\"" + authToken + "\",\"expires_in\":2678400,\"scope\":\"user:info user:check-access role:edit:syndesis-staging:! role:system:build-strategy-source:syndesis-staging\",\"token_type\":\"Bearer\"}"
+        )
+            .withHeader("Content-Type", "application/json")
+            .always();
+
+        KeycloakAuthenticationToken keycloakAuthenticationToken = new KeycloakAuthenticationToken(
```
KurtStam: This test is using the a GitHub token, why do we need to do anything keycloak?

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
```diff
@@ -62,13 +68,44 @@ public void before() throws IOException {
...
+            new SimpleKeycloakAccount(
+                new KeycloakPrincipal<RefreshableKeycloakSecurityContext>("testuser", null),
+                null,
+                new RefreshableKeycloakSecurityContext(
+                    null,
+                    null,
+                    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlzcyI6Imh0dHA6Ly9sb2NhbGhvc3Q6MTIzNC9hdXRoL3JlYWxtcy9zeW5kZXNpcy1pdCJ9.N0R7v55eiVXMzzhooYb1BStzyu_iP07wKNIO9z9dRMs",
```
KurtStam: What is this hash?

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
jimmidyson: The github token is retrieved from keycloak, an authenticated principal is required to do that.

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
jimmidyson: It's a dummy JWT we use for the test - you can inspect the contents at https://jwt.io/

core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -56,6 +68,65 @@ public static boolean isTokenExpired(String token) {
...
+            return verifier.getToken().getIssuer();
+        } catch (VerificationException e) {
+            throw SyndesisServerException.launderThrowable(e);
+        }
+    }
+
+    public static String getProviderToken(String providerId) {
+        String keycloakTokenAsString = getAuthenticationToken();
+
+        String issuer = getIssuer(keycloakTokenAsString);
```
rhuss: Question: The issuer of the keycloak token is always the same as the API Url to the keycloak server or is this just a convention ? And if it is a convention, one that can be taken for granted or something we need to document (e.g. as comment in the keycloak configmap) ?

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,171 @@
...
+     * Updates an existing git repository with the current version of project files.
+     *
+     * @param remoteGitRepoHttpUrl- the HTML (not ssh) url to a git repository
+     * @param repoName              - the name of the git repository
+     * @param message-              commit message
+     * @param files-                map of file paths along with their content
+     * @param credentials-          Git credentials, for example username/password, authToken, personal access token
+     * @return RevCommit, the commit info
+     */
+    public RevCommit updateFiles(String remoteGitRepoHttpUrl, String repoName, String message, Map<String, byte[]> files,
```
rhuss: The return value is never used, would make the method `void`

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,171 @@
...
+        for (Map.Entry<String, byte[]> entry : files.entrySet()) {
+            File file = new File(workingDir.toString(), entry.getKey());
+            if (!file.getParentFile().exists() && !file.getParentFile().mkdirs()) {
+                throw new IOException("Cannot create directory " + file.getParentFile());
+            }
+            Files.write(file.toPath(), entry.getValue());
+        }
+    }
+
+    private RevCommit commitAndPush(Git git, String message, UsernamePasswordCredentialsProvider credentials)
```
rhuss: Return value is never used, what about changing to `void` ?

github/src/main/java/io/syndesis/git/GitWorkflow.java
```diff
@@ -0,0 +1,171 @@
...
+     * the files to it.
+     *
+     * @param remoteGitRepoHttpUrl- the HTML (not ssh) url to a git repository
+     * @param repoName              - the name of the git repository
+     * @param message-              commit message
+     * @param files-                map of file paths along with their content
+     * @param credentials-          Git credentials, for example username/password, authToken, personal access token
+     * @return RevCommit, the commit info
+     */
+    public RevCommit createFiles(String remoteGitRepoHttpUrl, String repoName, String message, Map<String, byte[]> files,
```
rhuss: The return value is never used, maybe make the method return `void` ?

core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -38,13 +49,14 @@ public static String getAuthenticationToken() {
         Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
         if (authentication == null) {
             throw new IllegalStateException("Cannot set authorization header because there is no authenticated principal");
-        } else if (!KeycloakAuthenticationToken.class.isAssignableFrom(authentication.getClass())) {
+        }
```
rhuss: I think this method should be split into two methods as they serve completely different use cases:

* `getAuthenticationTokenForController` is called by the integration controllers which do the status updates (e.g. updating the files in git). This method looks up the token from a ThreadLocal (the first four lines in this method here). The caller of this method will never have an SecurityContextHolder from a web request set, so when the caller calls this method and there is no such token, it should error immediately (leading to a clearer errror message)

* `getAuthenticationTokenFromSecurityContext` is called by everyone in course of a web request handler chain. It should never lookup in the (controller reserved) thread local variable. 

Actually, I would put these methods even on different classes (the controller related more closely to the controllers themselves)

rhuss: If you feel that's not related to this PR, that's true ;-) we can do it in a separate PR, too.

github/README.md
```diff
@@ -0,0 +1,90 @@
...
+http://syndesis.io/quickstart/#deployment-instructions
+
+## Keycloak
+
+With the information above a Keycloak provider is created which at login time issues an OAuthToken that
+will be set on a header in each request. For this token to be able to be used by git, it needs to have
+'repo' scope. This scope info is set when Keycloak interacts with Github. See the table below for a full
+list of scopes supported by Github [1] and [2].
+
+|Name|	Description|
```
rhuss: Isn't it too confusing to list here all scopes ? Wouldn't it be simpler to just say which scope we need and then refer for the full list to GitHub docs ?

Also I don't think that in the current setup Keycloak interacts with GitHub, it just holds the GitHub token and returns it to syndesis-rest, which then in turn uses it to contact GitHub. (right?)

rhuss:
@jimmidyson @KurtStam would love to merge the PR soonish. Do you have time to work on the PR comments? If not I can take this over.

core/src/main/java/io/syndesis/core/Tokens.java
jimmidyson: The issuer should always be the same as the external URL to the keycloak server.

core/src/main/java/io/syndesis/core/Tokens.java
jimmidyson: Yeah I'd prefer to do that in a separate PR, it requires a bit of thought (oh no!) because this method is actually called in the git workflow, which doesn't know if it's getting the token from the session or thread local.

core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -56,6 +68,65 @@ public static boolean isTokenExpired(String token) {
...
+    private static String getIssuer(String token) {
+        TokenVerifier verifier = TokenVerifier.create(token);
+        try {
+            return verifier.getToken().getIssuer();
+        } catch (VerificationException e) {
+            throw SyndesisServerException.launderThrowable(e);
+        }
+    }
+
+    public static String getProviderToken(String providerId) {
```
jimmidyson: Done.

github/src/main/java/io/syndesis/git/GitWorkflow.java
jimmidyson: Done.

github/src/main/java/io/syndesis/git/GitWorkflow.java
jimmidyson: Done.

github/README.md
jimmidyson: Updated.

jimmidyson:
@rhuss Addressed your review comments.

rhuss:
[merge]

jimmidyson:
This means we can drop the github token-rp from deployment templates btw... I'll do that now

PR #470, opened 2017-07-28
[WIP] Integration runtime updates + refactoring

rhuss:


rhuss:
Need to fix some tests still.

jimmidyson:
@rhuss Fingers crossed the tests are now fixed up properly, not totally happy how the conversion between API model and integration runtime model is happening (marshal API model and then unmarshal to integration runtime model???) but that can be looked at separately I guess.

PR #480, opened 2017-08-01
fix(credentials): OAuth data needs to be applied at Connection creation

### LINKED ISSUE #477: Apply OAuth credential data at the point of connection creation
See the comment on https://github.com/syndesisio/syndesis-ui/issues/556#issuecomment-319096425 connectionId is not known by the UI at the point OAuth flow is initiated. We should store OAuth credential data in client side cookie store and apply not on the reception of OAuth tokens but at the point the OAuth capable connection is created.

This opens an issue of doing two OAuth flows in parallel (currently we distinguish them by connectionId).

zregvart:
This applies the OAuth data for a Connection at the point of Connection
creation. Previously a created connection was needed to initiate the
OAuth dance, this is no longer true.

Fixes #477


zregvart:
Same as in #486, @jimmidyson @iocanel @chirino would you mind reviewing this?

zregvart:
Fixing the Codacy issues first

PR #481, opened 2017-08-01
feat: Precompute atlasmap inspections for java objects and serve them as static resources.  Closes #466

### LINKED ISSUE #466: Precompute atlasmap inspections at build time and provide a service for them
Atlasmap now has a maven-plugin that can generate the inspections at build time.
Use that to precompute all the inspections and create a rest service that the UI can use to access those precomputed values.

chirino:


chirino:
Once merged, https://github.com/syndesisio/syndesis-openshift-templates/pull/65 will expose the new service via an openshift route.

chirino:
retest this please

syndesis-maven-plugin/src/main/java/io/syndesis/maven/GenerateMapperInspectionsMojo.java
```diff
@@ -0,0 +1,128 @@
...
+        try {
+
+            Resource resource = new Resource();
+            resource.setDirectory(outputDir.getCanonicalPath());
+            project.addResource(resource);
+
+            HashSet<File> generated = new HashSet<>();
+
+            ReadApiClientData reader = new ReadApiClientData();
+            List<ModelData<?>> modelList = reader.readDataFromFile("io/syndesis/dao/deployment.json");
```
rhuss: wouldn't it make sense to provide an explicit list to configure for the data types to extract ? And that already in the upstream plugin. 

I'm not sure that extracting the shapes from demo data is sufficient and reliable. Would prefer something like:

```xml
<configuration>
   <types>
       <twitter>
            <type>twitterj.Status</type>
             <type>...</type>
        </twitter>
        <saleforce>
             .....
        </salesforce>
   </types>
</configuration>
```

It's a bit tricky to setup this kind of configuration for Maven (but it works when using sth. like `Map<String, ArrayList>` as type for a `types` configuration field, see also [here](http://stackoverflow.com/questions/38628399/using-map-of-maps-as-maven-plugin-parameters/38642613) , but I could help after my PTO (next week). 

In the meantime, I'm ok with this as a temporary solution (but please open issue for the refactoring if you agree that we should couple the generation so tightly to the content of a file which can change anytime, or even vanish)

chirino:
retest this please

syndesis-maven-plugin/src/main/java/io/syndesis/maven/GenerateMapperInspectionsMojo.java
chirino: I'm not sure I agree with that.  We would end up duplicating configuration and that would lead to increased chances of getting it wrong.  By using the deployments.json as the source of truth, the generated pre-computations will always be accurate.

chirino:
retest this please

syndesis-maven-plugin/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+
+      <!--
+          Don't have time to resolve all the dependency issues, but it works
+          so disable all these plugin checks since it's only used at build time.
+      -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: surely we don't want to skip the dependency checks? perhaps add exceptions if versions are overridden, or redeclare dependencies with correct versions/scopes?

syndesis-maven-plugin/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>com.ning.maven.plugins</groupId>
+        <artifactId>maven-dependency-versions-check-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: surely we don't want to skip the dependency checks? perhaps add exceptions if versions are overridden, or redeclare dependencies with correct versions/scopes?

syndesis-maven-plugin/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.basepom.maven</groupId>
+        <artifactId>duplicate-finder-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: exceptions can be easily added when appropriate see https://github.com/syndesisio/syndesis-rest/blob/cc99fd13d77cb162da32bcc248a4e6db53476001/pom.xml#L786-L802

syndesis-maven-plugin/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>com.hubspot.maven.plugins</groupId>
+        <artifactId>dependency-management-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: surely we don't want to skip the dependency checks? perhaps add exceptions if versions are overridden, or redeclare dependencies with correct versions/scopes?

syndesis-maven-plugin/pom.xml
chirino: yes I do want to skip all of these.  The pom will bloat with a ton of config to add exceptions without actually providing any benefit since I'm not planning to resolve any of the exception.  It's much quicker and less config bloat to just disable the checks.

syndesis-maven-plugin/pom.xml
chirino: yes we do.

syndesis-maven-plugin/pom.xml
chirino: yes we do.

syndesis-maven-plugin/pom.xml
chirino: yes we do.

syndesis-maven-plugin/src/main/java/io/syndesis/maven/GenerateMapperInspectionsMojo.java
chirino: BTW that's not demo data.  Thats the official list of connectors that we are going to support right?  AFAIK there is no other way to add connectors into the system.

PR #483, opened 2017-08-02
fix(project): Dependency issues

zregvart:
This reinstates the dependency plugin checks. Also of note fixes
incorrect version of jackson-annotations (resolved 2.8.0, rest of the
jackson dependencies are at 2.8.8).

A number of jersey dependencies were excluded as they are not needed.

`syndesis-maven-plugin` module was placed before the `runtime` module
in order to be built in proper order.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -167,10 +167,10 @@
     <module>openshift</module>
     <module>project-generator</module>
     <module>rest</module>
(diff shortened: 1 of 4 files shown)


zregvart:
@chirino can you have a look, thanks 👍

PR #487, opened 2017-08-03
bug: Using wrong version in generated project.  Closes atlasmap/issues/144

chirino:


jludvice:
Closes atlasmap/atlasmap#144

PR #493, opened 2017-08-04
refactor: Introduce IntegrationRevision, IntegrationSpec and Integrat…

iocanel:
…ionStatus.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -26,4 +26,4 @@ coverage:
 comment:
   layout: "header, diff, tree, sunburst"
   require_changes: true
-   
+
--- a/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
+++ b/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
@@ -1,4 +1,4 @@
-/**
+ /**
  * Copyright (C) 2016 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,14 +17,17 @@ package io.syndesis.controllers.integration;
 
 import java.io.IOException;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
@@ -36,6 +39,8 @@ import io.syndesis.model.ChangeEvent;
 import io.syndesis.model.Kind;
 import io.syndesis.model.integration.Integration;
 
+import io.syndesis.model.integration.IntegrationRevision;
+import io.syndesis.model.integration.IntegrationState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -51,7 +56,9 @@ public class IntegrationController {
 
     private final DataManager dataManager;
     private final EventBus eventBus;
-    private final ConcurrentHashMap<Integration.Status, StatusChangeHandlerProvider.StatusChangeHandler> handlers = new ConcurrentHashMap<>();
+    private final Map<IntegrationState, StatusChangeHandler> draftHandlers;
+    private final Map<IntegrationState, StatusChangeHandler> deployedHandlers;
+
     private final Set<String> scheduledChecks = new HashSet<>();
     private ExecutorService executor;
     private ScheduledExecutorService scheduler;
@@ -59,14 +66,23 @@ public class IntegrationController {
     private static final long SCHEDULE_INTERVAL_IN_SECONDS = 60;
 
     @Autowired
-    public IntegrationController(DataManager dataManager, EventBus eventBus, StatusChangeHandlerProvider handlerFactory) {
+    public IntegrationController(DataManager dataManager, EventBus eventBus, StatusChangeHandlerProvider handlerProvider) {
         this.dataManager = dataManager;
         this.eventBus = eventBus;
-        for (StatusChangeHandlerProvider.StatusChangeHandler handler : handlerFactory.getStatusChangeHandlers()) {
-            for (Integration.Status status : handler.getTriggerStatuses()) {
-                this.handlers.put(status, handler);
-            }
-        }
+        this.draftHandlers = new HashMap<>();
+        this.deployedHandlers = new HashMap<>();
+
+        handlerProvider.getDraftStatusChangeHandlers().forEach(h ->
+           draftHandlers.putAll(h.getTriggerStatuses()
+               .stream()
+               .collect(Collectors.toMap(t -> t, t -> h)))
+        );
+
+        handlerProvider.getDeployedStatusChangeHandlers().forEach(h ->
+            deployedHandlers.putAll(h.getTriggerStatuses()
+                .stream()
+                .collect(Collectors.toMap(t -> t, i -> h)))
+        );
     }
 
     @PostConstruct
@@ -76,6 +92,13 @@ public class IntegrationController {
         scanIntegrationsForWork();
 
         eventBus.subscribe("integration-controller", getChangeEventSubscription());
+    }
+
+    @PreDestroy
+    public void stop() {
+        eventBus.unsubscribe("integration-controller");
+        scheduler.shutdownNow();
+        executor.shutdownNow();
     }
 
     private EventBus.Subscription getChangeEventSubscription() {
@@ -105,20 +128,25 @@ public class IntegrationController {
         executor.execute(() -> {
             Integration integration = dataManager.fetch(Integration.class, id);
             if( integration!=null ) {
-                String scheduledKey = getIntegrationMarkerKey(integration);
-                // Don't start check is already a check is running
-                if (!scheduledChecks.contains(scheduledKey)) {
-                    checkIntegrationStatus(integration);
-                }
+
+                integration.getDraftRevision().ifPresent(r -> {
+                    String scheduledKey = getIntegrationMarkerKey(integration, r);
+                    // Don't start check is already a check is running
+                    if (!scheduledChecks.contains(scheduledKey)) {
+                        checkIntegrationStatus(integration);
+                    }
+                });
+
+                integration.getDeployedRevision().ifPresent(r -> {
+                    String scheduledKey = getIntegrationMarkerKey(integration, r);
+                    // Don't start check is already a check is running
+                    if (!scheduledChecks.contains(scheduledKey)) {
+                        checkIntegrationStatus(integration);
+                    }
+                });
+
             }
         });
-    }
-
-    @PreDestroy
-    public void stop() {
-        eventBus.unsubscribe("integration-controller");
-        scheduler.shutdownNow();
-        executor.shutdownNow();
     }
 
     private void scanIntegrationsForWork() {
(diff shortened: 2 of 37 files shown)


jimmidyson:
@iocanel Is there an epic/issue to connect this PR with?

.codecov.yml
```diff
@@ -18,12 +18,12 @@ coverage:
   range: "30...90"
   precision: 2
   round: down
-  status:
+  state:
```
jimmidyson: Think this highlights some issues around global search and replace? Before I review this can you please just double check that all `status` -> `state` replacements in this PR are valid (this one is not!)?

jimmidyson (changes requested):
Please verify all `status` -> `state` replacements are correct in this PR - some are definitely not.

rhuss:
@jimmidyson It's about the integration lifecycle handling (jist linked it to the epic), also described in https://github.com/syndesisio/syndesis-project/blob/master/proposals/044-integration-model.md

rhuss:
I'm giving it a review tomorrow, as @iocanel is now on PTO for the next two weeks.

iocanel:
@rhuss: don't let the fact that I am on PTO discourage you from reviewing. I'll apply requests asap.

rhuss:
@iocanel ;-) I have to get the filter stuff flying today, but after this review of this PR will be on my plate.

model/src/main/java/io/syndesis/model/integration/IntegrationRevision.java
```diff
@@ -0,0 +1,110 @@
...
+            .currentState(state)
+            .currentMessage(message)
+            .build();
+    }
+
+    default IntegrationRevision withTargetState(IntegrationState state) {
+        return new IntegrationRevision.Builder().createFrom(this).targetState(state).build();
+    }
+
+    default IntegrationRevision.Builder newVersionBuilder() {
```
rhuss: `newIntegrationRevisionBuilder()` ?

model/src/main/java/io/syndesis/model/integration/IntegrationRevision.java
```diff
@@ -0,0 +1,110 @@
...
+    Integer getParentVersion();
+
+
+    IntegrationSpec getSpec();
+
+    /**
+     * The desired state of the revision.
+     * @return
+     */
+    IntegrationState getTargetState();
```
rhuss: I just had the chance to look around the way how Kubernetes / OpenShift does the resource management. They are using both a *status* and a *state* for describing the current situation of an resourxe: 

* `status` is a the upper object which holds information about the history of state changes (via its `conditions` subobject)
* `state` is enclosed in the status and contains the `currentState` and potentially `desiredState` (but this seems to be removed after being reached ?)

Actually wdyth to encapsulate the overall state related information (targetState (mayb rename back to desiredState ?), currentState, currentMessage, targetMessage, ..) into a `IntegrationStatus` object, too ? 

This helps people with a k8s background to better match the concepts (and might even help if we ever want to move to a real custom resource in k8s sometimes, but this is not something important for now).

rhuss (commented):
Just did a quick review over the integration model objects (not the logic yet). 

In order to start the discussion now, I submit this review early (with more comments to come later).

Only some minor question, also around how we might map status / state better to e.g. openshift concepts.

controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
```diff
@@ -51,22 +56,33 @@
...
+        this.draftHandlers = new HashMap<>();
+        this.deployedHandlers = new HashMap<>();
+
+        handlerProvider.getDraftStatusChangeHandlers().forEach(h ->
+           draftHandlers.putAll(h.getTriggerStatuses()
+               .stream()
+               .collect(Collectors.toMap(t -> t, t -> h)))
+        );
+
+        handlerProvider.getDeployedStatusChangeHandlers().forEach(h ->
```
rhuss: status2handler mapping func could be moved in a single method as it duplicated here.

controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
```diff
@@ -105,22 +128,27 @@ private void checkIntegrationStatusIfNotAlreadyInProgress(String id) {
...
+
+                integration.getDraftRevision().ifPresent(r -> {
+                    String scheduledKey = getIntegrationMarkerKey(integration, r);
+                    // Don't start check is already a check is running
+                    if (!scheduledChecks.contains(scheduledKey)) {
+                        checkIntegrationStatus(integration);
+                    }
+                });
+
+                integration.getDeployedRevision().ifPresent(r -> {
```
rhuss: duplicate code, please extract to single method.

model/src/main/java/io/syndesis/model/integration/IntegrationState.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * The {@link IntegrationRevision} is deployed but in an error state.
+     */
+    Error,
+
+    /**
+     * The {@link IntegrationRevision} is in peding state. (Desired != Actual).
+     */
+    Pending,
```
rhuss: `Pending` must not be an explicit state but is always calculated.

model/src/main/java/io/syndesis/model/integration/Integration.java
```diff
@@ -34,39 +34,45 @@
...
 
-    Optional<String> getDescription();
+    Optional<String> getToken();
 
     Optional<String> getGitRepo();
 
-    Optional<Status> getDesiredStatus();
-
-    Optional<Status> getCurrentStatus();
+    List<User> getUsers();
```
iocanel: I don't know. Users are not introduced in this PR. They were just a chunk of code moved to an other place.

model/src/main/java/io/syndesis/model/integration/IntegrationRevision.java
iocanel: This sounds good to me

model/src/main/java/io/syndesis/model/integration/IntegrationRevision.java
iocanel: Yeap, why not!

controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
```diff
@@ -67,29 +71,45 @@
...
+    /**
+     *
+     * @param integration
+     * @param revision
+     * @return
+     */
     @Override
-    public StatusUpdate execute(Integration integration) {
+    public StatusUpdate execute(Integration integration, IntegrationRevision revision) {
+        //TODO: This is clearly wrong, we need something better than that.
```
rhuss: Might it make sense that we add a max or last version counter directly on the integration ?

controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
```diff
@@ -134,101 +162,132 @@ private void checkIntegrationStatus(Integration integration) {
...
+            ? integration.getDraftRevision().get().withCurrentState(update.getState(), update.getStatusMessage())
+            : null;
+
+        //We need to iterate all revisions and update the status where there revision number matches.
+        List<IntegrationRevision> updatedRevisions = integration.getRevisions().stream()
+            .map(r -> r.getVersion().orElse(0).equals(revision.getVersion().equals(0))
+                ? r.withCurrentState(update.getState(), update.getStatusMessage())
+                : r)
+            .collect(Collectors.toList());
+
```
rhuss: What I'm missing here is, when a draft integration gets deployed, it should be added to the list of deployedRevisions. Could be that I missed that, but I thought this would be the correct place here ?

controllers/src/main/java/io/syndesis/controllers/integration/online/DeactivateHandler.java
```diff
@@ -64,11 +72,11 @@ public StatusUpdate execute(Integration integration) {
             }
         }
 
-        Integration.Status currentStatus = openShiftService.isScaled(deployment)
-            ? Integration.Status.Deactivated
-                : Integration.Status.Pending;
+        IntegrationState currentStatus = openShiftService.isScaled(deployment)
+            ? IntegrationState.Undeployed
```
rhuss: A deactivate handler should only put an integration to `deactivate`, never undeploy it. Also is must not set the current state to pending (but leave is untouched if the precondition is not matched). 

Was is currently missing here is the actual connection to the runtime to activel suspend it. This is related to the camel health check story (see syndesis board).

controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
```diff
@@ -46,7 +50,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ActivateHandler implements StatusChangeHandlerProvider.StatusChangeHandler {
+public class ActivateHandler implements StatusChangeHandler {
```
rhuss: Actually, this is the "DeployHandler" and should be renamed accordingly (refering to the proposal). What is missing is the "ResumeHandler" (opposite of "SuspendHandler").

controllers/src/main/java/io/syndesis/controllers/integration/online/DeactivateHandler.java
```diff
@@ -21,13 +21,16 @@
...
+import io.syndesis.controllers.integration.StatusUpdate;
 import io.syndesis.core.Tokens;
 import io.syndesis.model.integration.Integration;
+import io.syndesis.model.integration.IntegrationRevision;
+import io.syndesis.model.integration.IntegrationState;
 import io.syndesis.openshift.OpenShiftDeployment;
 import io.syndesis.openshift.OpenShiftService;
 
-public class DeactivateHandler implements StatusChangeHandlerProvider.StatusChangeHandler {
+public class DeactivateHandler implements StatusChangeHandler {
```
rhuss: "SuspendHandler"

controllers/src/main/java/io/syndesis/controllers/integration/online/DeleteHandler.java
```diff
@@ -18,13 +18,16 @@
...
+import io.syndesis.controllers.integration.StatusUpdate;
 import io.syndesis.core.Tokens;
 import io.syndesis.model.integration.Integration;
+import io.syndesis.model.integration.IntegrationRevision;
+import io.syndesis.model.integration.IntegrationState;
 import io.syndesis.openshift.OpenShiftDeployment;
 import io.syndesis.openshift.OpenShiftService;
 
-public class DeleteHandler implements StatusChangeHandlerProvider.StatusChangeHandler {
+public class DeleteHandler implements StatusChangeHandler {
```
rhuss: "UndeployHandler"

controllers/src/main/java/io/syndesis/controllers/integration/online/DeleteHandler.java
```diff
@@ -33,27 +36,34 @@
...
             .token(token)
             .build();
 
-        Integration.Status currentStatus = !openShiftService.exists(deployment)
+        IntegrationState currentStatus = !openShiftService.exists(deployment)
             || openShiftService.delete(deployment)
-            ? Integration.Status.Deleted
-            : Integration.Status.Pending;
+            ? IntegrationState.Undeployed
+            : IntegrationState.Pending;
```
rhuss: No "Pending" status, just leave either the currentStatus != targetStatus or set current status to "Error".

rhuss (changes requested):
Review, part 2

Looks good in general, however we need to hook it up to the UI in the next step to get even a better feeling about.

I would address the minor model changes and the more important comments in this PR first and then merge ASAP.

rhuss:
Superseded by #585

jimmidyson:
@iocanel @rhuss reopened this after last week's discussion with @gashcrumb to keep the API model the same for now, just introduce revisions for the UI to have history to display/manage. Do you think you can update this PR to reflect those changes as discussed?

jimmidyson:
@iocanel @gashcrumb Do we need to get this updated and merged so we can get the integration history page hooked up? That is what we discussed last week iirc, right?

iocanel:
@jimmidyson: No, its still the old thing.

PR #494, opened 2017-08-07
refactor(credentials): Cookie path set to `/`, added support for `X-Forwarded-Origin` header

zregvart:
Found it a bit cleaner to overwrite cookie than to manage lifecycle on
same cookie name but on two different paths (`.../connections` and
`.../callback`).
Also for local development behind ng middleware proxy added support for
`X-Forwarded-Origin` header.


rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
```diff
@@ -44,4 +60,58 @@ public static URI apiBase(final HttpServletRequest httpRequest) {
...
+        }
+
+        final String[] parts = origin.split(";");
+
+        String hostnameAndPort = null;
+        String scheme = null;
+        if (parts != null) {
+            for (final String part : parts) {
+                final int idx = part.indexOf('=');
+                final String name = part.substring(0, idx);
```
rhuss: What if idx == -1 because the header value is malformed ?

rhuss: Maybe return also an empty optional then

rest/src/test/java/io/syndesis/rest/v1/util/UrlsTest.java
```diff
@@ -42,4 +54,22 @@ public void shouldComputeBasePath() {
...
+
+    @Test
+    public void shouldDetermineCurrentUriFromRequestAndOriginHeader() {
+        final HttpServletRequest request = mock(HttpServletRequest.class);
+        when(request.getRequestURL()).thenReturn(new StringBuffer("https://syndesis.io/api/v1/resource/subresource"));
+        when(request.getHeader(Urls.ORIGIN_HEADER)).thenReturn("for=127.0.0.1;host=localhost:4200;proto=https");
+
+        assertThat(Urls.currentUri(request))
+            .isEqualTo(URI.create("https://localhost:4200/api/v1/resource/subresource"));
+    }
```
rhuss: Maybe a test for `parseOrigin` (with some faulty input) would be good also.

rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
```diff
@@ -44,4 +60,58 @@ public static URI apiBase(final HttpServletRequest httpRequest) {
         return "/" + parts[1] + "/" + parts[2] + "/";
     }
 
+    /* default */ static URI currentUri(final HttpServletRequest httpRequest) {
+        // in development we have Angular CLI proxying to HA proxy (openshift
+        // router) that proxies to the syndesis-rest; we need the origin
+        // hostname/port/scheme/
+        final String origin = httpRequest.getHeader(ORIGIN_HEADER);
```
rhuss: One question for my understanding: Who is setting ORIGIN_HEADER header ?

zregvart: The Angular CLI has a middleware proxy configured through `proxy.conf.js`: https://github.com/syndesisio/syndesis-ui/blob/master/proxy.conf.js#L10 -- it's a header that's set by the development environment for the UI.

rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
```diff
@@ -17,17 +17,33 @@
...
 
 public final class Urls {
 
+    /* default */ static final String ORIGIN_HEADER = "X-Forwarded-Origin";
+
     private Urls() {
         // Utility class, only static methods
     }
 
+    public static URI absoluteTo(final HttpServletRequest httpRequest, final URI url) {
```
jimmidyson: Does it make more sense to use URL rather than URI? We're always going to be working with URLs here right, rather than URIs?

rest/src/test/java/io/syndesis/rest/v1/util/UrlsTest.java
jimmidyson: Fuzzing ftw!

rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
zregvart: I don't think so URLs [are evil](https://docs.oracle.com/javase/8/docs/api/java/net/URL.html#equals-java.lang.Object-)

    Since hosts comparison requires name resolution, this operation is a blocking operation.

zregvart:
@rhuss @jimmidyson I've amended the commit (sorry, habit, only [Urls.java](https://github.com/syndesisio/syndesis-rest/pull/494/files#diff-483815cfed4bc1fb7b09fdd12c9014c7) and [UrlsTest.java](https://github.com/syndesisio/syndesis-rest/pull/494/files#diff-ff13ae85a01853670b74a2c996c3dc61) were changed), can you have another look? thanks 👍

rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
jimmidyson: Woah that's awful. Suggestion revoked.

PR #495, opened 2017-08-08
chore(validation): Adds unique name validator for Connections

### LINKED ISSUE #476: Validate connection name on creation
Connection names are supposed to be unique so the endpoint that creates new connections needs to validate a connection with the same name doesn't already exist.

zregvart:
This adds validation of Connection name uniqueness. The validation is
implemented as bean validation @UniqueProperty annotation (can be
reused).

Fixes #476


zregvart:
Would like to add constraint violation with the path to the property, will be better that way.

dao/src/main/resources/META-INF/validation.xml
```diff
@@ -0,0 +1,24 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<validation-config xmlns="http://jboss.org/xml/ns/javax/validation/configuration" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://jboss.org/xml/ns/javax/validation/configuration validation-configuration-1.1.xsd" version="1.1">
+
+  <constraint-mapping>META-INF/validation/unique-constraint.xml</constraint-mapping>
+
+</validation-config>
```
jimmidyson: nit: new line at end of file! This is serious... ;)

dao/src/test/java/io/syndesis/dao/DataManagerTest.java
```diff
@@ -51,9 +56,8 @@ public void setup() {
     @Test
     public void getConnectors() {
         ListResult<Connector> connectors = dataManager.fetchAll(Connector.class);
-        for (Connector connector : connectors.getItems()) {
-            System.out.print(connector.getId().get() + ",");
-        }
+        assertThat(connectors.getItems().stream().map(Connector::getId).map(Optional::get))
+            .containsExactly("gmail", "github", "ftp", "facebook", "linkedin", "salesforce", "timer", "jms", "day-trade", "twitter", "servicenow", "http", "trade-insight");
```
jimmidyson: Is this list going to be annoying to maintain as we add connectors or is this just a list of connectors specifically configured for the test?

dao/src/main/resources/META-INF/validation.xml
zregvart: Humble apologies, will amend 😳

dao/src/test/java/io/syndesis/dao/DataManagerTest.java
zregvart: I truly hope so, but slightly better than System.out's before that. Also these tests only test Infinispan cache 😞

zregvart: Aha and the list comes from `deployment.json`, so yes this will fail every time that changes.

PR #499, opened 2017-08-09
New resource for validation

### LINKED ISSUE #476: Validate connection name on creation

zregvart:
**Based on #495, please review it and merge before this**

Adds a new connection sub-resource (`validation`) that can be used to
perform validation of the soon-to-be created connection resource.

Full endpoint of the resource is `/api/v1/connections/validation` and
for connection payloads sent via `POST` method it will respond with
status `200` and 0-length body if the connection passes validation; or
status `409` and body containing a set of objects containing `property`
(what property) and `message` (validation message) properties.

Fixes #476

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/operations/Validating.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/operations/Validating.java
@@ -1,0 +1,63 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.rest.v1.operations;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.validation.ConstraintViolation;
+import javax.validation.Validator;
+import javax.validation.constraints.NotNull;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import io.syndesis.model.WithId;
+import io.syndesis.model.validation.AllValidations;
+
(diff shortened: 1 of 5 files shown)


akieling:
How is this PR going? It would be great if we could resolve it soon.

rest/src/main/java/io/syndesis/rest/v1/operations/Validating.java
```diff
@@ -0,0 +1,63 @@
...
+
+    Validator getValidator();
+
+    @POST
+    @Path(value = "/validation")
+    @Consumes("application/json")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiResponses({//
+        @ApiResponse(code = 200, message = "All validations pass"), //
+        @ApiResponse(code = 409, message = "Found violations in validation", responseContainer = "Set",
```
jimmidyson: Now I think about it, I think that 400 would be a better response as this is potentially not only testing for uniqueness (where 409 would make perfect sense).

runtime/src/test/java/io/syndesis/runtime/ConnectionsITCase.java
```diff
@@ -0,0 +1,65 @@
...
+        assertThat(got.getBody()).hasSize(1);
+    }
+
+    @Test
+    public void shouldDetermineValidityForValidConnections() {
+        @SuppressWarnings("PMD.CloseResource")
+        final Connection connection = new Connection.Builder().name("Test connection").build();
+
+        final ResponseEntity<List<Violation>> got = post("/api/v1/connections/validation", connection, RESPONSE_TYPE,
+            tokenRule.validToken(), HttpStatus.OK);
```
jimmidyson: Needs to be 204

runtime/src/test/java/io/syndesis/runtime/ConnectionsITCase.java
```diff
@@ -0,0 +1,65 @@
...
+        dataManager.create(connection);
+    }
+
+    @Test
+    public void shouldDetermineValidityForInvalidConnections() {
+        @SuppressWarnings("PMD.CloseResource")
+        final Connection connection = new Connection.Builder().name("Existing connection").build();
+
+        final ResponseEntity<List<Violation>> got = post("/api/v1/connections/validation", connection, RESPONSE_TYPE,
+            tokenRule.validToken(), HttpStatus.CONFLICT);
```
jimmidyson: Should be 400.

zregvart:
> Tests need fixing please.

Totally forgot, I'm going to fix the tests and squash.

rhuss:
I wonder how semantically the _validation_ here and the _verification_ of a connection is related. Technically _validation_ happens on syndesis-rest (the same for every connection), whereas _verification_ happens within a specific connector (which not only checks whether a connection is functional, but also whether the combination of parameters is allowed and syntactically correct).

zregvart:
@rhuss are you suggesting these two should be merged? I would prefer that the validation means that the entity makes sense syntactically, and verification that it makes sense for the runtime.

zregvart:
retest this please

rhuss:
@zregvart I'm fine with that, probably just wanted to mention the overlap.

Issue #451, opened 2017-07-21
stepKind not correct for Basic Filter


kahboom:
lol I don't know if selected the correct labels for this issue, but all seemed appropriate...

Reaching a 500 internal server error when trying to save an integration with the basic filter `configuredProperties`. See screenshots below.

Upon closer inspection of the [filter proposal](https://github.com/syndesisio/syndesis-project/blob/master/proposals/002-filter-data.md#domain), my best guess is that the `stepKind` has not been updated to be `basic-filter`. We have reserved the `stepKind` of `filter` for the advanced filter and `basic-filter` for the basic filter. Please note that we cannot use the same `stepKind` for both basic and advanced filters, as this property is used to determine which view to render.

<img width="1680" alt="screenshot 2017-07-20 22 30 36" src="https://user-images.githubusercontent.com/3844502/28447588-9d370fd0-6d9f-11e7-9b86-b0a1ef104e55.png">
<img width="1680" alt="screenshot 2017-07-20 22 31 35" src="https://user-images.githubusercontent.com/3844502/28447591-9ebe863a-6d9f-11e7-8205-a4a49008afcc.png">
<img width="1680" alt="screenshot 2017-07-20 22 32 53" src="https://user-images.githubusercontent.com/3844502/28447595-a251e134-6d9f-11e7-8636-609f2e53290c.png">
<img width="1679" alt="screenshot 2017-07-20 22 35 26" src="https://user-images.githubusercontent.com/3844502/28447596-a4d06cfa-6d9f-11e7-9f6a-fded700b66e7.png">
<img width="1083" alt="screenshot 2017-07-20 22 36 19" src="https://user-images.githubusercontent.com/3844502/28447600-a88313e8-6d9f-11e7-8144-b8822eac0890.png">

jimmidyson:
There shouldn't be a new step kind imo, as the proposal states. This is just a configured filter, either with `simple` field populated for basic filter or `rules` populated for advanced filters.

The view to render needs to be derived from these property values. Technically why can't that happen?

Fyi this shouldn't return a 500, but a 422 (unprocessable entity) with the appropriate validation error message to the client.

gashcrumb:
We don't have code in place to interpret what's in configured properties
when you pick a step, it's doable I guess but then you can't roundtrip
between the two views anyways.  So then they feel like two totally
different things.

jimmidyson:
[...]

PR #501, opened 2017-08-09
refactor(rule-filter): Adapt data format of a basic filter step according to the UI needs.

rhuss:
It's an ugly change. sorry. Instead of the typed interface
configuredProperties is used which contains values as serialized JSON strings.

Its change that way to make it happen and serious efforts should be undertaken to come to a more typed interface.

For possible options see https://github.com/syndesisio/syndesis-rest/issues/451#issuecomment-321249672

### DIFF
--- a/model/src/main/java/io/syndesis/model/filter/RuleFilterStep.java
+++ b/model/src/main/java/io/syndesis/model/filter/RuleFilterStep.java
@@ -15,11 +15,15 @@
  */
 package io.syndesis.model.filter;
 
+import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.syndesis.core.Json;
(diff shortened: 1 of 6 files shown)


model/src/main/java/io/syndesis/model/filter/RuleFilterStep.java
```diff
@@ -28,31 +32,48 @@
...
+        if (getConfiguredProperties().isPresent()) {
+            Map<String, String> props = getConfiguredProperties().get();
 
-    /**
-     * List of rules to combine.
-     */
-    List<FilterRule> getRules();
+            FilterPredicate predicate = getPredicate(props.get("predicate"));
+            List<FilterRule> rules = extractRules(props.get("rules"));
+            if (rules != null) {
```
jimmidyson: Or empty?

jimmidyson:
Will need to squash before merging.

rhuss:
> Minor nit but otherwise LGTM

well, doesn't look good to me but sometimes you have to bite the bullet ;-]

rhuss:
@jimmidyson a small approve, please ;-) !

rhuss:
hope you don't my second self-merge today (even should I rot in hell ;-)

[merge]

rhuss:
ok, still a minor issue to fix.

PR #504, opened 2017-08-10
fix(oauth): Fixes cookie path and JSON deserialization

### LINKED ISSUE #503: Not setting the Cookie and OAuth1 issues
At some point the Cookie used for client side persistence lost `path` parameter, which bound the cookie to the current path.

Also the `@JsonDeserialize` and the `converter` registered for `org.springframework.social.oauth1.OAuthToken` don't seem to make much difference when de-serializing the flow state from cookie.

zregvart:
Reverts to a _conservative_ version of cookie (with the Path, HttpOnly
and Secure parameters set), the cookie should traverse back and forth
when set from UI and received on OAuth callback.

Adds Jackson module specific to
`org.springframework.social.oauth1.OAuthToken` deserialization.

Fixes #503

### DIFF
--- a/credential/pom.xml
+++ b/credential/pom.xml
@@ -92,6 +92,11 @@
 
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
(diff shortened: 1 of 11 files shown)


zregvart:
> LGTM on a first look. will need to dive into the logic later, too ;-)

thanks!

I think we need rethink some of this client side persistance, it couples the UI to the backend a bit heavy handedly, also management of the state is spread over UI and backend so it could be difficult to maintain

PR #505, opened 2017-08-10
fix(rule-filter): Silly condition bug

rhuss:
that's the result of hectic PRs with no patience for waiting on reviews ;-(


rhuss:
[merge]

PR #512, opened 2017-08-11
fix(github): Correctly parse github provider token from keycloak

jimmidyson:


core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -106,25 +119,24 @@ public static String fetchProviderTokenFromKeycloak(String providerId) {
...
+                } catch (IOException e) {
+                    throw SyndesisServerException.launderThrowable(e);
+                }
+            case GITHUB:
+                Optional<NameValuePair> accessToken = URLEncodedUtils.parse(responseBody, StandardCharsets.UTF_8)
+                    .stream()
+                    .filter((nameValuePair) -> nameValuePair.getName().equalsIgnoreCase("access_token")).findFirst();
+                return accessToken.orElseThrow(IllegalStateException::new).getValue();
+            default:
+                throw new IllegalArgumentException("Unknown provider type: " + provider);
```
rhuss: nit: would call it "Unsupported" as it is known because we have an enum for it. Also, why not putting the extraction directly on to the enum ? would avoid the switch and might it make clearer.

rhuss:
The error I got was a 401  and unmarshal error from egit because it expected JSON.

I debugged into it and the error message was a text string like "Invalid JWS token, only 1 segment present". But I forgot to record the full message (before the liveness check killed my pod).

zregvart:
There is a nifty utility in `spring-web` that you can use also so we have one less dependency:

    MultiValueMap<String, String> params = UriComponentsBuilder.fromHttpUrl(...).build().getQueryParams()

jimmidyson:
Yeah we need to update the rest configmap to drop https://github.com/syndesisio/syndesis-openshift-templates/blob/master/generator/03-syndesis-rest.yml.mustache#L259-L260

core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -119,24 +136,11 @@ public static String fetchProviderTokenFromKeycloak(TokenProvider provider) {
...
-                    throw SyndesisServerException.launderThrowable(e);
-                }
-            case GITHUB:
-                Optional<NameValuePair> accessToken = URLEncodedUtils.parse(responseBody, StandardCharsets.UTF_8)
-                    .stream()
-                    .filter((nameValuePair) -> nameValuePair.getName().equalsIgnoreCase("access_token")).findFirst();
-                return accessToken.orElseThrow(IllegalStateException::new).getValue();
-            default:
-                throw new IllegalArgumentException("Unknown provider type: " + provider);
+        if (provider != TokenProvider.OPENSHIFT && provider != TokenProvider.GITHUB) {
```
rhuss: tbh I would even ommit this check and let the enum itself throw the IllegalArgumentException in apply() if it is not supported.

jimmidyson: Sorry being a bit slow this morning ;)

core/src/main/java/io/syndesis/core/Tokens.java
```diff
@@ -31,9 +33,34 @@
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilder;
 import java.io.IOException;
+import java.util.Locale;
+import java.util.function.UnaryOperator;
 
 public final class Tokens {
 
+    public enum TokenProvider implements UnaryOperator<String> {
+        GITHUB {
```
rhuss: @jimmidyson You implemented GITHUB and OPENSHIFT exactly in the opposite wat (openshift has json, github query params)

jimmidyson: Aaagggghhh sorry, copy-paste grrr.

jimmidyson: Fixed

zregvart:
retest this please

PR #515, opened 2017-08-11
feat(credentials): Mark connections as derived

### LINKED ISSUE #513: Identifier needed for connections that have been configured using an OAuth flow
As far as as I know there's no difference between a connection that was created by filling in the form in the create connection wizard vs. going through an OAuth flow.  For the 'edit' screen to display the right interface we need an attribute on the connection that the UI can use.

zregvart:
This adds a flag `derived` to Connections, derived Connections are those
that did not get their `configuredProperties` from user input but from
other means, such as OAuth dance performed by Credentials. Credentials
will also set this flag to `true` for Connections that pass successfully
through the OAuth acquisition flow and are saved at that point.

Once set the flag remains set until the UI chooses to reset it to
`false`, which it might if we ever decide to offer editing derived
Connections.

Fixes #513

### DIFF
--- a/credential/src/main/java/io/syndesis/credential/Credentials.java
+++ b/credential/src/main/java/io/syndesis/credential/Credentials.java
@@ -65,7 +65,11 @@ public final class Credentials {
     public Connection apply(final Connection updatedConnection, final CredentialFlowState flowState) {
         final CredentialProvider credentialProvider = providerFrom(flowState);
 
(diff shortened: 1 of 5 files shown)


zregvart:
> Looks good, but you probably need to update deployment.json, too, as derived is not marked as an optional field (?)

I thought as it was `boolean` (primitive type) it would be initialized to `false`, so I would leave it as is.

rhuss:
good point. Let it merge then.

Issue #418, opened 2017-07-12
Generated integration project should optionally use productized FIS bom


chirino:
The project generator should support switching to using productized FIS boms in the generated pom.xml files.

rhuss:
The benefit I guess would be to have the same version of camel libraries across all integrations, right? 

For the connectors however the versions are currently coming out of the DB, so I wonder how these would fit in the picture (and also of course, how to update connectors for deployed integrations).

chirino:
Right. Perhaps we should avoid adding connector versions in the DB and add those to the bom??
This goes hand in hand with how do we patch ipaas integrations.

rhuss:
I would keep this issue in mind, but first check that we get our update process straight. I guess an integration then must have least then a reference to the bom version and must have a way to either update the BOM via UI (admin area) or some automatic update which runs over all deployed integrations needs to be implemented.

chirino:
This should also include the FMP config in the pom.xml

lburgazzoli:
Would something like this:

```java
@Value.Immutable
public interface GenerateProjectRequest {
    enum Layout {
        FIS,
        OSS
    }

    Layout getProjectLayout();
    ...
}
```

be ok to discriminate the project layout ?

rhuss:
I think that's perfectly fine. Maybe we call it the Layout 'REDHAT' and 'SYNDESIS' or so, because it might be used also for other dependencies used in the generated pom.xml (which for the RedHat case must only use productised dependencies, including connectors).

The layout then could be selected by a simple configuration option.

What will be also interesting is how to configure the appropriate repositories (with authentication for the redhat case ?)

lburgazzoli:
A maybe better alternative would be to use spring-boot profiles so i.e. we can set the path of the templates and additional stuffs in each profile

lburgazzoli:
Should we close this as it has been merged ?

gashcrumb:
+1

lburgazzoli:
Please reopen if needed

PR #519, opened 2017-08-14
WIP - Generated integration project should optionally use productized FIS bom #418

lburgazzoli:
This PR is not to be merged but I've opened it to gather some feedback.

Now project-generator can be configured with an ``template override path`` which is the location were overriding templates are stored. In addition you can now list a number of additional resources that can be included i.e. the ``redhat`` profile includes some fabric8 bits like ``src/main/fabric8``

### DIFF
--- a/project-generator/src/main/java/io/syndesis/project/converter/DefaultProjectGenerator.java
+++ b/project-generator/src/main/java/io/syndesis/project/converter/DefaultProjectGenerator.java
@@ -17,8 +17,13 @@ package io.syndesis.project.converter;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
@@ -36,8 +41,10 @@ import io.syndesis.integration.model.Flow;
 import io.syndesis.integration.model.SyndesisHelpers;
 import io.syndesis.integration.model.SyndesisModel;
 import io.syndesis.integration.model.steps.Endpoint;
+import io.syndesis.integration.support.Strings;
 import io.syndesis.model.integration.Integration;
 import io.syndesis.model.integration.Step;
+import io.syndesis.project.converter.ProjectGeneratorProperties.Templates;
 import io.syndesis.project.converter.visitor.GeneratorContext;
 import io.syndesis.project.converter.visitor.StepVisitor;
 import io.syndesis.project.converter.visitor.StepVisitorContext;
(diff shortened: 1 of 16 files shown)


rhuss:
thanks @lburgazzoli I'm going to review it asap (a lot of people are on PTO, I will be on public holiday and PTO on Tuesday and Friday). So bear a bit with me .... ;-)

lburgazzoli:
@rhuss PTO for me tomorrow so no worries :)

jimmidyson:
@lburgazzoli Some test failures on Jenkins and Circle - can review once those are fixed up.

project-generator/src/main/resources/application-redhat.properties
```diff
@@ -0,0 +1,8 @@
+
+generator.templates.override-path = redhat
+
+generator.templates.additional-resources[0].sourece = deployment.yml
```
rhuss: typo ?

project-generator/src/main/java/io/syndesis/project/converter/DefaultProjectGenerator.java
```diff
@@ -104,9 +123,31 @@ public DefaultProjectGenerator(ConnectorCatalog connectorCatalog, ProjectGenerat
         });
 
         Map<String, byte[]> contents = new HashMap<>();
+
+        for (Templates.Resource additionalResource : generatorProperties.getTemplates().getAdditionalResources()) {
+            String overridePath = generatorProperties.getTemplates().getOverridePath();
+            URL resource = null;
+
+            if (!Strings.isEmpty(overridePath)) {
+                resource = getClass().getResource("templates/" + overridePath + "/" + additionalResource.getSource());
```
rhuss: What if overridePath is provided but the resource could not be found ? The overridePath should be at least included in the error message, but maybe already throw an exception here.

lburgazzoli: The logic is that it checks first if the listed resource is part of the "overridden" resources then it search in the default path so the exception is thrown only if the resource is not found in any of the known path, may be we should log a debug level which path is used ?

rhuss: ok, got it. I think it would be sufficient if we add the overridePath to an IOException if thrown. I think for debugging its enough if we add a comment top-level on the generated files for what they are for (e.g. via XML / properties comments in the templates)

lburgazzoli:
Reworked a little bit

project-generator/src/main/java/io/syndesis/project/converter/DefaultProjectGenerator.java
```diff
@@ -104,9 +124,36 @@ public DefaultProjectGenerator(ConnectorCatalog connectorCatalog, ProjectGenerat
...
+            }
+            if (resource == null) {
+                resource = getClass().getResource("templates/" + additionalResource.getSource());
+            }
+            if (resource == null) {
+                throw new IllegalArgumentException(
```
rhuss: thx ! :)

lburgazzoli:
@rhuss should we provide any customization for the deployment.yml (if it is needed) ?

lburgazzoli:
@chirino do you mind having a look to see if it copes with the requirements ?

zregvart:
@lburgazzoli there is a [dependency check issue reported](https://jenkins-syndesis-ci.b6ff.rh-idev.openshiftapps.com/blue/organizations/jenkins/syndesis-rest/detail/PR-519/6/pipeline/#step-20-log-1353), you can run the checks locally with: `./mvnw -Pchecks clean install`.

PR #522, opened 2017-08-15
fix(github): Set commit author name and email from github retrieved info

### LINKED ISSUE #518: bug(github): Email and User of git commits are not properly set
The generated and checked in files in GitHub do not contain a proper user or email name (but only a single '?') mark. Wonder whether this should be a configuration option, too ?

jimmidyson:
Fixes #518

### DIFF
--- a/controllers/pom.xml
+++ b/controllers/pom.xml
@@ -101,6 +101,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.eclipse.mylyn.github</groupId>
+      <artifactId>org.eclipse.egit.github.core</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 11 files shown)


jimmidyson:
The test failure highlight something I've not thought about: what if the user doesn't have a name configured (that is allowed in GitHub profile) and/or their email address is not public? How about a default global setting for fallback name/email?

rhuss:
A user has always a GitHub user id, or ? I would use this as fallback to the name. For the email, I wonder whether it is mandatory to have is set for git.

But having a fallback in the global settings GitHub tab (which we add anyway for setting up the OAuth params) makes sense to me, too. At least for the email.

zregvart:
There is one error in the integration test:

```
[ERROR] testGetApiUser(io.syndesis.github.GitHubServiceITCase)  Time elapsed: 0.196 s  <<< FAILURE!
java.lang.AssertionError:  Expecting actual not to be null
	at io.syndesis.github.GitHubServiceITCase.testGetApiUser(GitHubServiceITCase.java:124)
[ERROR] testProjectCommit(io.syndesis.github.GitHubServiceITCase)  Time elapsed: 1.884 s  <<< ERROR!

java.lang.IllegalArgumentException: Name of PersonIdent must not be null.
	at io.syndesis.github.GitHubServiceITCase.testProjectCommit(GitHubServiceITCase.java:174)
```

rhuss:
[merge]

rhuss:
sorry, my bad that I merged by accident. Going to have a look soonish.

PR #530, opened 2017-08-21
build(jenkins): Use new style stage blocks to remove deprecation warnings

jimmidyson:


zregvart:
While we're at it would it make sense to use Maven wrapper `./mvnw` instead of Maven tool?

jimmidyson:
@zregvart Separate PR? I do agree though.

PR #532, opened 2017-08-22
chore(build): Use Maven wrapper

zregvart:
Let's use Maven wrapper instead of `mvn` from Maven image. We can still
keep the Maven image.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,3 @@
-def mavenVersion='3.3.9'
-
 properties([
     buildDiscarder(logRotator(numToKeepStr: '5', artifactNumToKeepStr: '5'))
 ])
@@ -14,7 +12,7 @@ node {
     slave {
         withOpenshift {
             withMaven(
-                mavenImage: "maven:${mavenVersion}",
+                mavenImage: "openjdk:8",
                 envVars: [
                     containerEnvVar(key:'GITHUB_OAUTH_CLIENT_ID', value: "${env.GITHUB_OAUTH_CLIENT_ID}"),
                     containerEnvVar(key:'GITHUB_OAUTH_CLIENT_SECRET', value: "${env.GITHUB_OAUTH_CLIENT_SECRET}")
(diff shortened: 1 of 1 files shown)


zregvart:
This won't build until https://github.com/syndesisio/syndesis-pipeline-library/pull/18 is merged.

iocanel:
retest this please

Jenkinsfile
```diff
@@ -14,7 +12,7 @@ node {
     slave {
         withOpenshift {
             withMaven(
-                mavenImage: "maven:${mavenVersion}",
+                mavenImage: "openjdk:8",
```
iocanel: I guess, if we are going to move down that way we need to rename 'mavenImage' to just 'image'

iocanel: at some point

zregvart:
Again, needs the latest fix of the pipeline https://github.com/syndesisio/syndesis-pipeline-library/pull/21

zregvart:
Not until we fix https://github.com/syndesisio/syndesis-pipeline-library/issues/23

KurtStam:
How does this compare to https://github.com/syndesisio/syndesis-rest/pull/721 and what are we trying to solve?

zregvart:
I think #721 defines a Maven image with some artifacts pre-cached to speed up the build, see 

https://github.com/syndesisio/syndesis-pipeline-library/blob/bea370f930e20b244db33687c35b9b3026460cbc/vars/withMaven.groovy#L32

https://github.com/syndesisio/syndesis-ci/blob/c3a1454e5f50b290165c4fcf2e38567e8a765df3/m2-bc.yml

https://github.com/syndesisio/syndesis-jenkins/tree/8ac2cb7d9d57c9b39548ccdc0ddcb0f3916f126e/configuration/jobs/m2-repo-builder

Whereas this would make use of Maven Wrapper to run the build instead of Maven from the image. Would be nice to have same Maven version run locally and on CI, but we also control the Maven version on CI, so I'm not sure if this is still needed. Food for thought...

PR #533, opened 2017-08-22
chore(validation): Refine validations

zregvart:
Adds a `error` property to `Violation` that holds the name of the
constraint annotation.

Changed `ConstraintViolationExceptionMapper` and `Validating` to make
the responses from `/validation` endpoint and from other endpoints that
are performing validations are unified.

Removed unneeded `@Size` constraint from `WithName` - now that the
strings in JSON payloads are trimmed to `null` `@NotNull` is the only
constraint needed.

Refined validation messages.


zregvart:
@akieling just as a heads up with this the JSON returned for validation will contain `message` (human readable text), `property` (path to the property) and `error` (the kind of error), see https://github.com/zregvart/syndesis-rest/blob/08695621b0e65e91b4f6aa117421e0559d5c717c/rest/src/main/java/io/syndesis/rest/v1/operations/Violation.java

runtime/src/main/resources/messages.properties
```diff
@@ -14,5 +14,9 @@
 # limitations under the License.
 #
 
-io.syndesis.credential.RelativeUri.message = Given URI value must be relative
-io.syndesis.model.validation.UniqueProperty.message = Value `${validatedValue[value]}` for property `{value}` is not unique
+io.syndesis.credential.RelativeUri.message          = Given URI value must be relative
```
rhuss: Not deep in bean validation, but can the value of the URI included in the error message ? Or is this done by default ? The more context, the better ...

runtime/src/main/resources/messages.properties
```diff
@@ -14,5 +14,9 @@
...
+io.syndesis.model.validation.UniqueProperty.message = Value is not unique
```
rhuss: Same here (would be could to see the actual non-unique value in the error message)

runtime/src/test/java/io/syndesis/runtime/ConnectionsITCase.java
```diff
@@ -80,7 +83,8 @@ public void shouldDetermineValidityForInvalidConnections() {
         final ResponseEntity<List<Violation>> got = post("/api/v1/connections/validation", connection, RESPONSE_TYPE,
             tokenRule.validToken(), HttpStatus.BAD_REQUEST);
 
-        assertThat(got.getBody()).hasSize(1);
+        assertThat(got.getBody()).containsExactly(
+            new Violation.Builder().property("name").error("UniqueProperty").message("Value is not unique").build());
```
rhuss: see above, would love to see the actual value here in the error message.

rhuss:
retest this please

rhuss:
ok, it's a test failure from another PR:

```
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 13.576 s <<< FAILURE! - in io.syndesis.github.GitHubServiceITCase
[ERROR] testProjectCommit(io.syndesis.github.GitHubServiceITCase)  Time elapsed: 3.531 s  <<< ERROR!
java.lang.IllegalArgumentException: E-mail address of PersonIdent must not be null.
	at io.syndesis.github.GitHubServiceITCase.testProjectCommit(GitHubServiceITCase.java:172)
```

zregvart:
Squashed and rebased on top of #542 so look at that first 😹

zregvart:
> my favorite counter example is Oracles 'invalid column name error' without saying what the invalid column name is (nice when you have a select going over 20 columns or so).

So you haven't dealt with Informix? "Syntax error" for every kind of syntactical error, the only way to get to the issue is to look at the log on server.

Also as a bonus DB2: "The value of a host variable in the EXECUTE or OPEN statement is too large for its corresponding use", for any value of any column being too large.

runtime/src/main/resources/messages.properties
```diff
@@ -14,5 +14,9 @@
 # limitations under the License.
 #
 
-io.syndesis.credential.RelativeUri.message = Given URI value must be relative
-io.syndesis.model.validation.UniqueProperty.message = Value `${validatedValue[value]}` for property `{value}` is not unique
+io.syndesis.credential.RelativeUri.message          = Given URI value ('{value}') must be relative
+io.syndesis.model.validation.UniqueProperty.message = Value '${nonUnique}' is not unique
```
rhuss: Not an BeanValidation connoisseur at all, but why are parameters sometimes referenced with `${...}` and sometimes with `{...}` ? Reminds me so to the JSF fiasco (`#{...}` anyone ?)

zregvart:
> Not an BeanValidation connoisseur at all, but why are parameters sometimes referenced with ${...} and sometimes with {...} ? Reminds me so to the JSF fiasco (#{...} anyone ?)

`{}` ones are standard ones, `${}` ones are Hibernate Validator ones (de facto standard), not sure why they can't be `{}`, could be that `${}` are EL expressions. Either way test proves the correctness.

PR #535, opened 2017-08-23
chore(metadata): Expand action with metadata

zregvart:
This adds the `ActionMetadata` and `ActionMetadata.Step` as proposed in
syndesisio/syndesis-project#72.


model/src/main/java/io/syndesis/model/connection/ActionMetadata.java
```diff
@@ -0,0 +1,62 @@
...
+            addPropertyDefinitionStep(stepBuilder.build());
+
+            return this;
+        }
+    }
+
+    @Value.Immutable
+    @JsonDeserialize(builder = Step.Builder.class)
+    @ImmutablesStyle
+    interface Step extends WithName, WithProperties {
```
rhuss: We already have `Step` as part of an integration definition. Maybe we can call it differently here like in `ActionDefinitionStep` ?

model/src/main/java/io/syndesis/model/connection/ActionMetadata.java
```diff
@@ -0,0 +1,62 @@
...
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(builder = ActionMetadata.Builder.class)
+public interface ActionMetadata {
+
+    final class Builder extends ImmutableActionMetadata.Builder {
+        // make ImmutableActionMetadata.Builder accessible
+
+        public Builder withStep(final String name, final String description, final Consumer<Step.Builder> stepConfigurator) {
```
rhuss: `withActionDefinitionStep` ?

model/src/main/java/io/syndesis/model/connection/Action.java
```diff
@@ -54,4 +54,6 @@ default Action withId(String id) {
     class Builder extends ImmutableAction.Builder {
     }
 
+    ActionMetadata metadata();
```
rhuss: Could we call it something different like `metadata` ? Maybe `definition` ?

The reason why I ask is, that there is still the idea that we change our resource model to a more openshift/kubernetes like format, where we mimic the k8s resource format here. The reasoning for this is, that this keeps the door somewhat open, that in a future version we could easily migrate to K8s _customer resource definitions_ (CRD) which e.g. use secrets etc. directly. 

It's not that far, but `metadata` is a _reserved_ word for such a resoruce definition which all look like:

```yaml
apiVersion:
kind:
metadata:
   ...
spec:
   ...
status:
  ....
```

Also I think that _definition_ is a bit more specific than _metadata_

zregvart:
> Can this be really a static one, or could it be that even the possible definition (fields) depend on a selected value from a previous step ?

Not sure, would opt for having it static in the first iteration, but the .../metadata endpoint allows it to be dynamic if needed.

rhuss:
> Not sure, would opt for having it static in the first iteration, but the .../metadata endpoint allows it to be dynamic if needed.

Ok, I'm fine with that for now.

rhuss:
@gashcrumb @zregvart when we merge this PR then this directly affects the UI because of the changed way of connector's properties.

@gashcrumb Is it fine for you todo the switch right now ? Or should we still hold on ?

rhuss:
@zregvart should we also update deployment.json to a new schema version so that the DB gets recreated during deployment ?

rhuss:
retest this please

gashcrumb:
@rhuss Yeah, I'm fine with this getting merged, I can deal with any breakage in the UI today.  We also definitely make sure stuff in the `deployment.json` matches too so we don't have stale test/example data, that can be a big thorn in my side :-)

rhuss:
@gashcrumb ok, lets get that out
@zregvart there are test failures

rhuss:
Test failure comes from another PR already merged:

```
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 13.576 s <<< FAILURE! - in io.syndesis.github.GitHubServiceITCase
[ERROR] testProjectCommit(io.syndesis.github.GitHubServiceITCase)  Time elapsed: 3.531 s  <<< ERROR!
java.lang.IllegalArgumentException: E-mail address of PersonIdent must not be null.
	at io.syndesis.github.GitHubServiceITCase.testProjectCommit(GitHubServiceITCase.java:172)
```

zregvart:
I think I need to create an integration test also, I think the handler is mapped to the wrong URL. (Will put a WIP label).

zregvart:
I think I changed the `deployment.json` to match this in https://github.com/syndesisio/syndesis-rest/pull/535/files#diff-42a113fed8dd9cf4d3605a145856da51

rhuss:
retest this please

zregvart:
I've rebased on current `master` and added integration test along with a slight refactoring and fixing the endpoint.

PR #538, opened 2017-08-23
fix(openshift): Removes consecutive dashes from...

### LINKED ISSUE #537: Better project name sanitization in OpenShiftService
From syndesisio/syndesis-ui#818 it appears that the name of the project was generated with consecutive `-` characters and that is not allowed with f8 k8s client.

zregvart:
... generated project name

Now any consecutive dashes are replaced with single dash. The Fabric8
Kubernetes Client with regex `[a-z0-9]+(?:[._-][a-z0-9]+)*` does not
allow for dash to be out dashed by another dash.

Fixes #537

### DIFF
--- a/core/src/main/java/io/syndesis/core/Names.java
+++ b/core/src/main/java/io/syndesis/core/Names.java
@@ -39,13 +39,21 @@ public final class Names {
      */
     public static String sanitize(String name) {
         String trimmed = name.length() > MAXIMUM_NAME_LENGTH ? name.substring(0, MAXIMUM_NAME_LENGTH) : name;
+
         return trimmed
             .replaceAll(SPACE, DASH)
             .replaceAll(INVALID_CHARACTER_REGEX, BLANK)
             .toLowerCase(Locale.US)
             .chars()
             .filter(i -> !String.valueOf(i).matches(INVALID_CHARACTER_REGEX))
-            .collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append)
+            .collect(StringBuilder::new,
+                (b, chr) -> {
+                    int lastChar = b.length() > 0 ? b.charAt(b.length() - 1) : -1;
+
+                    if (lastChar != '-' || chr != '-') {
+                        b.appendCodePoint(chr);
+                    }
+             }, StringBuilder::append)
             .toString();
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/syndesis/core/NamesTest.java
```diff
@@ -15,29 +15,42 @@
...
 
-            "aaa-big-and-small-zzz", "AaA BIG and Small ZzZ",
-        };
+    @Parameters
+    public static Iterable<Object[]> integrationToProjectNames() {
+        return Arrays.asList(pair("bla", "bla"),
+            pair("0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789",
+                "012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"),
+            pair("how-are-you", "how are you?"), //
+            pair("yet-sth-with-spaces", "yet sth  with !#ä spaceS"),
```
rhuss: would add a test with three and more spaces, too.

core/src/main/java/io/syndesis/core/Names.java
```diff
@@ -39,13 +39,22 @@ private Names() {
...
         return trimmed
             .replaceAll(SPACE, DASH)
             .replaceAll(INVALID_CHARACTER_REGEX, BLANK)
             .toLowerCase(Locale.US)
             .chars()
             .filter(i -> !String.valueOf(i).matches(INVALID_CHARACTER_REGEX))
-            .collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append)
+            .collect(StringBuilder::new,
+                (b, i) -> {
+                    if (i == '-' && last[0] == '-') {
```
rhuss: fine with that, but maybe a bit more concise as 

```java
if (i != '-' || last[0] != '-') {
  b.appendCodePoint(i);
  last[0] = i;
}
```

rhuss: Actually the trick to update an outer array feels a bit dirty ;-)

maybe an anonymous class with a member variable instead of a lambda would be a better fit ?

rhuss:
@zregvart According to the regexp in the error message, this kind of squeezing should happen for "." and "_", too.

zregvart:
> @zregvart According to the regexp in the error message, this kind of squeezing should happen for "." and "_", too.

We replace with only `'-'` so `'.'` and `'_'` don't appear in the sanitized name.

rhuss:
[merge]

PR #542, opened 2017-08-25
fix(github): generate e-mail address for users ...

zregvart:
...that elected not to show their e-mail address publicly

If a GitHub user has chosen not to display the e-mail address in public
then API will not return the e-mail address. For this case we need to
use a GitHub provided no-reply e-mail address.

Fixes #522

### DIFF
--- a/github/src/main/java/io/syndesis/github/GitHubServiceImpl.java
+++ b/github/src/main/java/io/syndesis/github/GitHubServiceImpl.java
@@ -16,6 +16,8 @@
 package io.syndesis.github;
 
 import java.io.IOException;
+import java.time.LocalDate;
+import java.time.ZoneId;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


github/src/main/java/io/syndesis/github/GitHubServiceImpl.java
```diff
@@ -79,7 +83,24 @@ public String getCloneURL(String repoName) throws IOException {
...
-        return userService.getUser();
+        final User user = userService.getUser();
+        // if the user did not elect to publicly display his e-mail address, e-mail will be null
+        // https://developer.github.com/v3/users/#get-a-single-user
+        // let's put a dummy e-mail address then, as it is needed for the commit
+        if (user.getEmail() == null) {
+            // users before 2017-07-18 have their no-reply e-mail addresses in the form
+            // username@users.noreply.github.com, and after that date
+            // id+username@users.noreply.github.com
+            // https://help.github.com/articles/about-commit-email-addresses/
```
rhuss: wow, cool finding. And also many thanks for commenting the details (although the urls might get stale sometimes, but really appreciate it)

PR #547, opened 2017-08-29
refactor(filter): Remove prefix for filter options

### LINKED ISSUE #517: bug(rule-filter): Paths contains wrong prefix
The rest endpoint which returns the possible options for a filter rule (wrt to the body) contains e.g. for twitter a "Status." prefix before the property (e.g. `Status.text`). Actually, this gets translated 1:1 to the generated simple expression. But instead of `Status.text` it must be `body.text`.

There can be two possible solutions:

* Either already use `body.text` for the auto suggestions in the filter rules (instead of `Status.text`)
* Translate automatically `<prefix>.<prop>` to `body.<prop>` when generating the simple expression. However, this becomes difficult if e.g. already well-known [simple](http://camel.apache.org/simple.html) prefixes like `in`, `headers`, `properties`, ... are used. 

Another, but a bit more involved solution would be, that in the options endpoint not only a single values path is returned but an object with two values: The one to display and autocomplete on (`Status.text`) and the one to be sent when the step is saved (`in.body.text` or `body.text`). @gashcrumb would this be too complicated to implement in the UI ? The drawback is that they would need to be translated back when editing the rule. hmmm.

So I tend to use the plain simple text (`body.user.screenName`) or even only without prefix at all `user.screenName` for the UI and then add the proper prefix when generating the rule.

E.g. a step would then have the following `configuredProperties`

```json
...
configuredProperties:
  predicated: "AND"
  pathPrefix: "Status"
  rules:
  - path: "user.screenName"
  - op: "equals"
  - value: "ro14nd"
```

This would be the most user-friendly variant (as you don't have to enter the common prefix "Status" at all). However, this would also close the door for 'advanced' expressions like filtering on the header or properties in this basic filter (one would need to use the advanced filter then). But considering we are targeting the citizen user here this might be not too much of a restriction (and even helps to keep it more simple) ?

rhuss:
This is not needed since only the latest output shape is used now for filtering.

Still: only java types are supported for the fields (see ClassInspector). This queries the atlasmap service internally.

Fixes #517.

### DIFF
--- a/core/src/test/java/io/syndesis/core/json/StringTrimmingJsonDeserializerTest.java
+++ b/core/src/test/java/io/syndesis/core/json/StringTrimmingJsonDeserializerTest.java
@@ -27,7 +27,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import org.junit.Test;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.*;
 
 public class StringTrimmingJsonDeserializerTest {
 
--- a/core/src/test/java/io/syndesis/core/json/StringTrimmingJsonDeserializerTest.java
+++ b/core/src/test/java/io/syndesis/core/json/StringTrimmingJsonDeserializerTest.java
@@ -27,7 +27,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import org.junit.Test;
 
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
 
 public class StringTrimmingJsonDeserializerTest {
 
--- a/inspector/pom.xml
+++ b/inspector/pom.xml
@@ -52,6 +52,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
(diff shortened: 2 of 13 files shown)


zregvart:
Seems that one of the tests is still on JSONPath syntax:
```
[ERROR] shouldDeserializeRuleFilterStep(io.syndesis.model.integration.StepDeserializerTest)  Time elapsed: 0.059 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<${body[.text} == 'antman' && ${header.kind]} =~ 'DC Comics'> but was:<${body[[body][text]} == 'antman' && ${body[header][kind]]} =~ 'DC Comics'>
	at io.syndesis.model.integration.StepDeserializerTest.shouldDeserializeRuleFilterStep(StepDeserializerTest.java:40)
```

core/src/test/java/io/syndesis/core/json/StringTrimmingJsonDeserializerTest.java
```diff
@@ -27,7 +27,7 @@
 
 import org.junit.Test;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.*;
```
zregvart: not sure this was intentionally included in the PR

rhuss: no, gonna rollback.

rhuss:
Sorry, messed up the tests (again was to eager with pushing this PR).

Gimme a second, will fix them.

rhuss:
@zregvart I think this PR is now good enough for a review. Sorry for the test malaise.

PR #548, opened 2017-08-29
chore(metadata): Adds support for dynamic action properties

zregvart:
Adds `/api/v1/connections/{connectionId}/actions/{actionId}/properties`
endpoint that responds to GET request and in the query parameters
expects any action property - value pairs that have been selected by the
user and offers suggestions as to values of other action properties.

The implementation delegates to `syndesis-verifier` and the endpoint
created in syndesision/syndesis-verifier#18[1] that in turn delegates
to Camel Metadata extension.

[1] https://github.com/syndesisio/syndesis-verifier/pull/18

### DIFF
--- a/model/src/main/java/io/syndesis/model/connection/Connector.java
+++ b/model/src/main/java/io/syndesis/model/connection/Connector.java
@@ -66,4 +66,9 @@ public interface Connector extends WithId<Connector>, WithName, WithProperties, 
     class Builder extends ImmutableConnector.Builder {
     }
 
+
+    default Optional<Action> actionById(String id) {
+        return getActions().stream().filter(a -> a.idEquals(id)).findFirst();
+    }
+
 }
--- a/model/src/test/java/io/syndesis/model/connection/ActionPropertySuggestionsTest.java
+++ b/model/src/test/java/io/syndesis/model/connection/ActionPropertySuggestionsTest.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.model.connection;
+
+import java.io.IOException;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.junit.Test;
+
+import static io.syndesis.model.connection.ActionPropertySuggestions.ActionPropertySuggestion.Builder.of;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ActionPropertySuggestionsTest {
+
+    private final static JsonNode JSON = parse("/action-property-suggestions.json");
+
+    private final static ObjectMapper MAPPER = new ObjectMapper();
+
+    private final static ActionPropertySuggestions SUGGESTIONS = new ActionPropertySuggestions.Builder()
+        .putValue("property1", of("1value1", "1displayValue1"), of("1value2", "1displayValue2"))
+        .putValue("property2", of("2value1", "2displayValue1"), of("2value2", "2displayValue2")).build();
+
+    @Test
+    public void shouldDeserialize() throws IOException {
+        final ActionPropertySuggestions deserialized = MAPPER.readerFor(ActionPropertySuggestions.class)
+            .<ActionPropertySuggestions>readValue(JSON);
+        assertThat(deserialized).isEqualTo(SUGGESTIONS);
+    }
+
+    @Test
+    public void shouldSerialize() {
+        final JsonNode serialized = MAPPER.valueToTree(SUGGESTIONS);
+        assertThat(serialized).isEqualTo(JSON);
+    }
+
+    private static JsonNode parse(final String path) {
+        try {
+            return new ObjectMapper().readTree(ActionPropertySuggestionsTest.class.getResourceAsStream(path));
+        } catch (final IOException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
+
+}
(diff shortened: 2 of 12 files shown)


dao/src/main/resources/io/syndesis/dao/demo-data.json
```diff
@@ -30,9 +30,10 @@
       "connectorId": "salesforce",
       "configuredProperties": {
         "loginUrl": "https://login.salesforce.com",
-        "clientId": "salesforce-client-id",
-        "clientSecret": "salesforce-client-secret",
-        "refreshToken": "salesforce-refresh-token"
+        "clientId": "3MVG9szVa2RxsqBZXHfqsW3hf9HQp_N6qdSmpjKMzSJaEL4UP161JlDkE32EigL82ra_jM1WuQgF4rYDgzL3u",
```
rhuss: Is this real data ? Shouldn't we avoid this for the demo data ?

model/src/main/java/io/syndesis/model/connection/Connector.java
```diff
@@ -66,4 +66,9 @@ default Builder builder() {
     class Builder extends ImmutableConnector.Builder {
     }
 
+
+    default Optional<Action> actionById(String id) {
+        return getActions().stream().filter(a -> a.getId().get().equals(id)).findFirst();
```
rhuss: short question (but probably the wrong place here): Why is an actionId an optional when it is mandatory ?

model/src/test/java/io/syndesis/model/connection/ActionPropertySuggestionsTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.junit.Test;
+
+import static io.syndesis.model.connection.ActionPropertySuggestions.ActionPropertySuggestion.Builder.of;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ActionPropertySuggestionsTest {
+
+    private final static String JSON = "{\"property1\":[{\"displayValue\":\"1displayValue1\",\"value\":\"1value1\"},{\"displayValue\":\"1displayValue2\",\"value\":\"1value2\"}],\"property2\":[{\"displayValue\":\"2displayValue1\",\"value\":\"2value1\"},{\"displayValue\":\"2displayValue2\",\"value\":\"2value2\"}]}";
```
rhuss: actually, I prefer to have this kind of test data in an external, pretty-printed resource file as its easier to update and understand.

model/src/test/java/io/syndesis/model/connection/ActionPropertySuggestionsTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final static ObjectMapper MAPPER = new ObjectMapper();
+
+    private ActionPropertySuggestions suggestions = new ActionPropertySuggestions.Builder()
+        .putValue("property1", of("1value1", "1displayValue1"), of("1value2", "1displayValue2"))
+        .putValue("property2", of("2value1", "2displayValue1"), of("2value2", "2displayValue2")).build();
+
+    @Test
+    public void shouldSerialize() throws JsonProcessingException {
+        assertThat(MAPPER.writerFor(ActionPropertySuggestions.class).writeValueAsString(suggestions)).isEqualTo(JSON);
```
rhuss: Test looks a bit fragile (as it depends how the mapper serializes the JSON to a string). I would prefer to use something like https://github.com/skyscreamer/JSONassert which is much robuster (and works with the suggestion from above)

rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionActionHandler.java
```diff
@@ -58,4 +84,38 @@ public ActionDefinition definition(@PathParam("id") @ApiParam(required = true,
...
+            .forEach(step -> step.getProperties().forEach((k, v) -> parameters.putIfAbsent(k, null)));
+
+        // lastly put all connection properties
+        parameters.putAll(connection.getConfiguredProperties());
+
+        final Client client = ClientBuilder.newClient();
+        final WebTarget target = client
+            .target(String.format("http://%s/api/v1/action/properties/%s", config.getService(), connectorId));
+
+        return target.request(MediaType.APPLICATION_JSON)
```
rhuss: I know this is the simplest way to get data from the syndesis-verifier (we should really check for a new name ;-) to the UI but it also tightly couples syndesis-ui to syndesis-verifier via this data type (ActionPropertySuggestions). The same problem we already have for the validation calls and which already leads to some strange constructs in the client (there is also an issue open for this to fix).

IMO it would be much better do decouple those (e.g. that the syndesis-verifier has not direct or indirect dependency on syndesis-rest) and let the verifier stay in charge of the model object it serves (and uses as request parameters).

Haven't yet looked in the verifier code, but the dependency chain (including 'copied' objects) should be 
`syndesis-ui --> syndesis-rest --> syndesis-verifier` and it should be unidirectional (i.e. no circles)

rhuss: Just have looked at the corresponding syndesis-verifier PR (https://github.com/syndesisio/syndesis-verifier/pull/18) and it seems that there is no directly coupling between object (but with the expected data format so that it can be serialized on both sides). 

I now just would wait on feedback from the UI what is the best way to consume the data and if there are changes required we can put the transformation into this here.

model/src/main/java/io/syndesis/model/connection/Connector.java
zregvart: Yeah, we have a couple of `Optional`'s in our model that could be revisited, also if it's optional that method should handle that.

zregvart:
This is a really bad PR, I've amended it with a few fixes, and I'm in the process of adding tests.

zregvart:
I've squashed and rebased on master to fix that compile error caused by #555 that was merged in a bit before this.

@KurtStam @rhuss @gashcrumb this adds the GET method as described in syndesisio/syndesis-project#72 for getting action property values values at runtime.

rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionActionHandler.java
```diff
@@ -58,4 +84,42 @@ public ActionDefinition definition(@PathParam("id") @ApiParam(required = true,
...
+        // put all action parameters with `null` values
+        action.getDefinition().getPropertyDefinitionSteps()
+            .forEach(step -> step.getProperties().forEach((k, v) -> parameters.putIfAbsent(k, null)));
+
+        // lastly put all connection properties
+        parameters.putAll(connection.getConfiguredProperties());
+
+        final Client client = createClient();
+        final WebTarget target = client
+            .target(String.format("http://%s/api/v1/action/properties/%s", config.getService(), connectorId));
```
rhuss: Wonder whether we should change to https at some point. Same for the actual verifier connection.
Opened https://github.com/syndesisio/syndesis-rest/issues/565 for this.

PR #553, opened 2017-08-30
fix(oauth): Adds more detailed status of OAuth ...

### LINKED ISSUE #514: Is there a cookie or session value that the rest API returns after an OAuth flow
When the user is redirected back to the UI there's no way currently (that I know of) that we can tell what state the OAuth flow is in, is it completed, successful, failed, etc.  The UI would need some way of figuring this out, maybe a value in the cookie that reflects the state of the oauth flow, or an API to verify the oauth state for the selected connector ID?  That way we can either just forward the user on to the next step or have whatever feedback mechanism is in the design (can't recall off the top of my head) working to inform the user.

zregvart:
...callback in the fragment

This would help the UI know if OAuth authorization succeeded or not. The
fragment of the given `returnUrl` is now ignored and replaced with a URL
encoded JSON that resembles something like:

    #%7B%22connectorId%22:%22test-provider%22,%22message%22:%22Successfully%20authorized%20Syndesis's%20access%22,%22status%22:%22SUCCESS%22%7D

(for successful authorization). Which translates to the following JSON:

    {
      "connectorId: "test-provider", // can be "twitter" or "salesforce"
      "message": "Successfully authorized Syndesis's access",
      "status": "SUCCESS"
    }

Status can also be "FAILURE" indication, well, failure in authorization,
the message offers a bit more detail.

Note(1): connectorId JSON property can be `null`, for instance if the
callback was invoked without `cred-` cookies, in that case there is no
flow to look for connectorId.

Note(2): The redirect might lead to the root of the application (i.e to
`/`) for the same reason.

Fixes #514

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
@@ -49,6 +49,17 @@ public final class Urls {
         }
     }
 
+    public static URI appHome(final HttpServletRequest httpRequest) {
+        final URI current = currentUri(httpRequest);
+
+        try {
+            return new URI(current.getScheme(), null, current.getHost(), current.getPort(), null, null, null);
+        } catch (final URISyntaxException e) {
+            throw new IllegalArgumentException(
+                "Unable to generate app home URI based on the current (`" + current + "`)", e);
+        }
+    }
+
     public static URI apiBase(final HttpServletRequest httpRequest) {
         final URI current = currentUri(httpRequest);
 
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/credential/CallbackStatus.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/credential/CallbackStatus.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


zregvart:
**This needs to be merged in coordination with UI!**

gashcrumb:
@zregvart do you wanna merge this and then I can sort it out today?

rest/src/main/java/io/syndesis/rest/v1/util/Urls.java
```diff
@@ -49,6 +49,17 @@ public static URI absoluteTo(final HttpServletRequest httpRequest, final URI url
         }
     }
 
+    public static URI appHome(final HttpServletRequest httpRequest) {
+        final URI current = currentUri(httpRequest);
+
+        try {
+            return new URI(current.getScheme(), null, current.getHost(), current.getPort(), null, null, null);
```
rhuss: That's a bit dangerous when the UI is not mapped to the root context. But I also don't see a better solution.

I wonder though whether the 'current' URI also depends on the SSL termination of the route. When the route terminates at the router, then it acts as a reverse proxy to the backend, which means you will probably see the internal URI address here (and hence a redirect won't work).  

Maybe we should put the home URL also in to the application.properties and let it fill in from the openshift tempalte ?

zregvart: I guess that would be much better, note that in #546 we've configured Undertow to to acknowledge `X-Forwarded-` headers set by nginx. And in OAuth (credentials) there are a number of places absolute URIs are computed.

PR #572, opened 2017-09-04
chore(tags): Set properties via tags

zregvart:
This allows properties being set via tags. Useful when the name of the
property is not known, but the tag is unique (enough) to use it to set
the property.

### DIFF
--- a/model/src/main/java/io/syndesis/model/connection/WithPropertiesBuilder.java
+++ b/model/src/main/java/io/syndesis/model/connection/WithPropertiesBuilder.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


jimmidyson:
Some test failure but can't quickly see exactly what it is... we should collect test logs for display in Jenkins.

model/src/main/java/io/syndesis/model/connection/WithPropertiesBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Sets property tagged with the given tag to the supplied value
+     *
+     * @param properties The properties
+     * @param tag The looked after tag
+     * @param value The value to set the property to
+     * @return The previous value of the property, if any
+     */
+    default T putOrRemoveconfiguredPropertyTaggedWith(final WithPropertiesBuilder<T> builder, final String tag,
```
jimmidyson: Typo: `configured` -> `Configured`

PR #573, opened 2017-09-04
refactor(credentials): Don't rely on explicit registration of CredentialProviders

### LINKED ISSUE #569: reset-db should clean also data for api/v1/connectors/twitter/credentials endpoint
When there are no settings/twitter secrets,

entpoint https://syndesis-qe.b6ff.rh-idev.openshiftapps.com/api/v1/connectors/twitter/credentials
returns empty json `{}`

once setup, this endpoint returns
```json
{"description":"twitter","icon":"twitter","label":"twitter","type":"OAUTH1"}
```

Then I call Reset db, and expect endpoint to return empty json again,
but it still returns 
```json
{"description":"twitter","icon":"twitter","label":"twitter","type":"OAUTH1"}
```

We can't proceed with e2e tests unless we're able to flush state of app!

CC @dsimansk @mmelko @lvydra

zregvart:
Based on #572 so take a look at that first (split to help the review)

This removes explicit registration of CredentialProviders and uses the
database as the source for configuring the CredentialProviders. Also
removes the magical auto-configuration via properties -
CredentialProviders will be configured only through the database so
there is no need for that.

Fixes #569


jimmidyson:
retest this please

credential/src/main/java/io/syndesis/credential/CredentialProviderRegistry.java
```diff
@@ -16,20 +16,54 @@
...
+
+        credentialProviderFactories = SpringFactoriesLoader
+            .loadFactories(CredentialProviderFactory.class, ClassUtils.getDefaultClassLoader()).stream()
+            .collect(Collectors.toMap(CredentialProviderFactory::id, Function.identity()));
     }
 
     @Override
     public CredentialProvider providerWithId(final String providerId) {
-        final CredentialProvider providerWithId = providers.get(providerId);
+        final Connector connector = dataManager.fetch(Connector.class, providerId);
```
jimmidyson: This seems odd to fetch the `Connector` using `providerId`. Is this method really better called `providerForConnector` and `providerId` -> `connectorId`?

jimmidyson: Ah the `providerId === credentialsId` always?

zregvart: Yeah I've stumbled on that myself, provider comes from Spring Social and we use Connector to name the same thing in this context. I could do a PR to switch all usages of `provider` to `connector`, would that be OK?

jimmidyson: Perhaps just raise an issue for it now, it's not urgent, just slightly confusing naming.

jimmidyson: Will merge this as-is anyway - thanks!

PR #574, opened 2017-09-04
fix(settings): Allows updating OAuth settings with empty or null values

### LINKED ISSUE #536: Can't remove oauth client credentials anymore
On staging clicking 'Remove' no longer clears stored client credentials, instead you get this error popup:

`Failed to delete settings: Response with status: 500 Internal Server Error for URL: https://localhost:4200/api/v1/setup/oauth-apps/twitter`

This used to work, note to clear the credentials we're currently setting the ID and secret to empty strings, so I guess this is a side-effect of trimming incoming JSON messages.

@zregvart @jludvice FYI...

zregvart:
Builds on #572 so have a look at that first

In the effort to remove OAuth configuration OAuth settings endpoint must
allow empty or null values. With trim-to-null empty values were
transformed to null and this did not allow the Connector's configured
properties to be updated. This builds on the PR#572 and allows setting
empty values for OAuth properties in the effort to remove those from
`configuredProperties`.

Fixes #536

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/setup/OAuthAppHandler.java
@@ -16,9 +16,7 @@
 package io.syndesis.rest.v1.handler.setup;
 
 import java.util.ArrayList;
-import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


zregvart:
Rebased on master

Issue #509, opened 2017-08-11
Canceling Twitter OAuth Flow gives error


rhuss:
When canceling the Twitter OAuth Flow via the "Cancel" button on the Twitter side, after the redirect the following error appears:

```json
{"errorCode":404,"userMsg":"Please check your request data","developerMsg":"Entity Not Found Exception Unable to finish OAuth authorization via callback, cannot find the OAuth flow state this callback request relates to"}
```

zregvart:
see #464 and #514

zregvart:
This should have been fixed with #577

Issue #464, opened 2017-07-26
Error handling in OAuth flows


zregvart:
We should have error handling in OAuth flows, the UI needs to be made aware of any errors and those need to be presented from the backend with messages that can be consumed by users.

zregvart:
Should be fixed with #577

PR #577, opened 2017-09-05
fix(credentials): Error handling in OAuth flows

### LINKED ISSUE #557: Clicking cancel in twitter during oauth flow causes a redirect back to the root of the syndesis app
Also I noticed the connectorId is 'null' instead, so unfortunately we lose all context of what the user was doing before starting the oauth flow.

zregvart:
This changes the way credential state cookies are handled so when the
flow is unsuccessful most recent state is considered to contain
providerId (connectorId) and returnUrl. That makes the UI receive proper
callback.

Fixes #557, #464, #509

### DIFF
--- a/credential/src/main/java/io/syndesis/credential/CredentialFlowStateHelper.java
+++ b/credential/src/main/java/io/syndesis/credential/CredentialFlowStateHelper.java
@@ -16,68 +16,41 @@
 package io.syndesis.credential;
 
 import java.util.Arrays;
-import java.util.Objects;
-import java.util.function.BiFunction;
-import java.util.stream.Stream;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import io.syndesis.credential.CredentialFlowState.Builder.Restorer;
+
+import org.apache.commons.lang3.ArrayUtils;
 
 final class CredentialFlowStateHelper {
-
-    private static final Logger LOG = LoggerFactory.getLogger(CredentialFlowStateHelper.class);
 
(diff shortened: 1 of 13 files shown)


credential/src/main/java/io/syndesis/credential/CredentialFlowStateHelper.java
```diff
@@ -16,68 +16,39 @@
...
-    }
-
-    /* default */ static Stream<CredentialFlowState> restoreFrom(
-        final BiFunction<javax.ws.rs.core.Cookie, Class<CredentialFlowState>, CredentialFlowState> restore,
-        final HttpServletRequest request, final HttpServletResponse response) {
+    /* default */ static Set<CredentialFlowState> restoreFrom(final Restorer restore,
+        final HttpServletRequest request) {
         final Cookie[] servletCookies = request.getCookies();
 
         if (servletCookies == null || servletCookies.length == 0) {
```
jimmidyson: Could use `ArrayUtils.isEmpty`

credential/src/main/java/io/syndesis/credential/OAuth1CredentialFlowState.java
```diff
@@ -86,6 +86,10 @@ default Type type() {
     default CredentialFlowState updateFrom(final HttpServletRequest request) {
         final String verifier = request.getParameter("oauth_verifier");
 
+        if (verifier == null || verifier.isEmpty()) {
```
jimmidyson: Could use `StringUtils.isEmpty`

credential/src/main/java/io/syndesis/credential/OAuth2CredentialFlowState.java
```diff
@@ -90,6 +90,10 @@ default Type type() {
     default CredentialFlowState updateFrom(final HttpServletRequest request) {
         final String code = request.getParameter("code");
 
+        if (code == null || code.isEmpty()) {
```
jimmidyson: Could use `StringUtils.isEmpty`

zregvart:
Rebased on #574 and incorporated review feedback

PR #580, opened 2017-09-07
chore(metadata): Metadata model changes

zregvart:
This is the first step into aligning the model with the latest version
of the metadata design proposal[1] to test the water and assess any UI
issues that might come of it.

[1] syndesisio/syndesis-project#72

### DIFF
--- a/dao/src/main/resources/io/syndesis/dao/demo-data.json
+++ b/dao/src/main/resources/io/syndesis/dao/demo-data.json
@@ -97,16 +97,18 @@
             "id": "1",
             "connectorId": "io.syndesis:twitter-mention-connector:latest",
             "description": "Connection from twitter when anyone mention you",
-            "inputDataShape": {
-                "kind" : "none"
-            },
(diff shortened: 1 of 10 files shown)


jimmidyson:
@gashcrumb @kahboom Do you know about these changes and what impact they'll have on the UI? Would like to merge ASAP to keep this moving, but not at the expense of breaking stuff for you unless you're ready to fix up pretty much immediately.

zregvart:
I was actually hoping we could merge this before I move on with the implementation of syndesisio/syndesis-project#72 to assess any UI breaking. AFAIK with `inputDataShape` and `outputDataShape` still present on `Action` (the same way `properties` are -- delegating to `ActionDefinition`) the UI should not be affected (this was requested in feedback by @gashcrumb on syndesisio/syndesis-project#72).
With UI feedback we can assess if we can deliver metadata/dynamic feature for TP1.

gashcrumb:
Yep, with that last change I think the UI should be good I think.

PR #582, opened 2017-09-08
chore(metadata): Align the implementation with the latest design document

zregvart:
This unifies two existing endpoints:

 - GET /api/v1/connections/{connectionId}/actions/{actionId}/definition
 - GET /api/v1/connections/{connectionId}/actions/{actionId}/properties

into a single endpoint:

 - POST /api/v1/connections/{connectionId}/actions/{actionId}

The existing action property suggestions are folded in without the
change on the `syndesis-verifier` and there is no support for data
shapes as there was no support in the previous implementation.

The new endpoint receives a key-value map of configured properties and
returns the action definition. It can receive no content, empty JSON
object (`{}`) or one or more action property values
(`{"prop":"value",...}`).

Every invocation of this endpoint results in the cascading invocation of
`syndesis-verifier`.


jimmidyson:
retest this please

PR #585, opened 2017-09-10
Integration Model revisited and updated

rhuss:
This work is based on https://github.com/syndesisio/syndesis-rest/pull/493 which has been previously updated with theses changes a week ago, but probably wiped out again with a with a forced push of https://github.com/syndesisio/syndesis-rest/pull/493/commits/7fdec31dd24e0d34dfe757ac61119d2a881e01f6.

An IntegrationRevision (which is part of the Integration) has now a forma
that ressembles that of K8s resources:

```yml 
spec:
  # Desired state
  state: "Published" 
status:
  # Current state
  state: "Draft"
  # Actual version of the revision
  version: 1
  # Message from the last state change
  message: "...."
```

Convenient method the the `curentState` and `desiredState` has been added
as methods to `IntegratioRevision`.

@iocanel @jimmidyson It would be awesome, if we can make some progress here. I will be busy with my talk for JavaZone till Thursday, but it would be awesome if we could have at least a review on this PR till then. I will try to be as responsive as possible.

thanks ...

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -26,4 +26,4 @@ coverage:
 comment:
   layout: "header, diff, tree, sunburst"
   require_changes: true
-   
+
--- a/controllers/src/main/java/io/syndesis/controllers/integration/DemoHandlerProvider.java
+++ b/controllers/src/main/java/io/syndesis/controllers/integration/DemoHandlerProvider.java
@@ -18,10 +18,13 @@ package io.syndesis.controllers.integration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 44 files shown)


jimmidyson:
We discussed this with @gashcrumb last week and decided that we need to keep the existing API model backwards compatible. What do you think? The problem with an invasive change like this at the moment is that it requires changes to the UI which I don't think we'll have time to do before release. Keeping old fields as deprecated sounds good to me.

gashcrumb:
Another option is to maybe make use of our versioning in the API path and start developing this at `/v2` maybe?

PR #587, opened 2017-09-11
chore(metadata): Limits dynamic action paramete...

### LINKED ISSUE #586: Let's limit dynamic action parameters and dynamic data shapes to tagged actions
For some actions it makes no sense to invoke the _verifier_. As it stands now, the `/api/v1/connections/{connectionId}/actions/{actionId}` endpoint invokes the _verifier_ in all cases.

So let's do in only for actions that are tagged with `dynamic` to filter out the cases where this is unnecessary.

zregvart:
...rs and dynamic data shapes to tagged actions

Adds tags to Action's and checks if the action is tagged with `dynamic`
to return either the definition contained in the database or the
definition enriched by invoking the _verifier_ to fetch additional
action property values or dynamic data shapes.

Fixes #586

### DIFF
--- a/model/src/main/java/io/syndesis/model/connection/Action.java
+++ b/model/src/main/java/io/syndesis/model/connection/Action.java
@@ -29,13 +29,14 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.syndesis.model.Kind;
 import io.syndesis.model.WithId;
 import io.syndesis.model.WithName;
+import io.syndesis.model.WithTags;
 
 import org.immutables.value.Value;
 
 @Value.Immutable
 @JsonDeserialize(builder = Action.Builder.class)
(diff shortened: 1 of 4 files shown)


jimmidyson:
We discussed this with @gashcrumb last week and decided that we need to keep the existing API model backwards compatible. What do you think? The problem with an invasive change like this at the moment is that it requires changes to the UI which I don't think we'll have time to do before release. Keeping old fields as deprecated sounds good to me.

zregvart:
Just to elaborate, we can use this `dynamic` tag on actions to turn on/off invocation of the _verifier_ for metadata. It also gives us the opportunity to hardcode values in the `enum` property of action definition and mark the action as non-dynamic and still have the UI working with multiple action property definition steps and _dynamic_ (sort of) action properties.

Issue #576, opened 2017-09-05
Update filter API to accept just an output data shape instead of the whole integration


gashcrumb:
So this can be merged:

https://github.com/syndesisio/syndesis-ui/pull/879

zregvart:
@igarashitm can you take a look at this, doesn't seem to big of a task at first glance, but if you have difficulties with it I can do it also, thanks 👍

igarashitm:
merged

PR #588, opened 2017-09-11
[#576] Update filter API to accept just an output data shape instead …

igarashitm:
…of the whole integration


zregvart:
nit: @igarashitm we use [conventional commits](https://syndesis.io/community/engineering_guidelines/#commit-messages-should-follow-the-conventional-commits-standardized-commit-message-format), so a commit title would look like `chore(filters): Update filter API to accept just an output data shape...` and in the commit message body you would end with `Fixes #576`

zregvart:
@gashcrumb this is what we want for #576, an API change? Or would you prefer to have both options available?

igarashitm:
OK fixed PMD errors and commit message

gashcrumb:
@zregvart yep, the current basic filter UI already passes only the data shape object when calling this API, so this should get that typeahead working again when there's a data shape available.

PR #593, opened 2017-09-13
Maintain the integration history without breaking the api.

iocanel:
This is a subset of https://github.com/syndesisio/syndesis-rest/pull/493, but including only changes that doesn't affect actual the API. Also, keeps behavioral changes to the bare minimum.


iocanel:
@gashcrumb: Now, the integration object contains a set of revisions (wich provide the historical data).

zregvart:
That PMD issue is failing the build:
```
[INFO] PMD Failure: io.syndesis.controllers.integration.online.ActivateHandler:82 Rule:PrematureDeclaration Priority:3 Avoid declaring a variable if it is unreferenced before a possible exit point..
```

jimmidyson:
@zregvart Yeah I'm fixing that now thanks

Issue #609, opened 2017-09-15
Generated projects do not include endpoint options


lburgazzoli:
If you generate a simple integration and you set values from the timer period and http-get/post url, the information is lost and not set on the steps:
 
```yaml
---
flows:
- steps:
  - kind: endpoint
    uri: periodic-timer:every
  - kind: endpoint
    uri: http-get
```

nor on the secrets:

```yaml
apiVersion: v1
data:
  application.properties: IwojVGh1IFNlcCAxNCAwOTozMDoyNiBVVEMgMjAxNwo=
kind: Secret
metadata:
  creationTimestamp: 2017-09-14T09:30:26Z
  name: syndesis-timer-14
  namespace: syndesis
  resourceVersion: "17738"
  selfLink: /api/v1/namespaces/syndesis/secrets/syndesis-timer-14
  uid: 578e00cd-992f-11e7-a04f-a207645835de
type: Opaque
```

lburgazzoli:
Can we close this ?

zregvart:
Let's close this and reopen or file new issues for any new bugs.

PR #612, opened 2017-09-15
Generated projects do not include endpoint options #609

lburgazzoli:


project-generator/src/main/java/io/syndesis/project/converter/visitor/EndpointStepVisitor.java
```diff
@@ -110,9 +107,9 @@ public EndpointStepVisitor(GeneratorContext generatorContext) {
         //
         final String connectorScheme = hasComponentOptions ? connection.getId().map(id -> camelConnectorPrefix + "-" + id).orElse(camelConnectorPrefix) : camelConnectorPrefix;
 
-        Map<String, String> secrets = connector.filterProperties(
+        Map<String, String> secrets = action.filterProperties(
             properties,
-            connector.isSecret(),
+            action.isSecret(),
```
zregvart: Do we need to check with connection if the property is secret also?
something like `connection.isSecret().or(action.isSecret())`

_Sidenote_
At first method named `is...` got me worried but then I understood that it returns a `Predicate`, wonder if anyone else thinks that we could have method handle here or name those `is...` methods returning `Predicate` differently, something like `test...`?

lburgazzoli: Yeah, will add such check.

Sidenote: 
yeah, there are now two variant of the isSecret, one that returns a predicate and one that can be used as method reference, I kept the predicate one for backward compatibility for the moment.

lburgazzoli: @zregvart should be fixed

PR #615, opened 2017-09-16
(sql-stored) setting the dynamic action config and setting the templa…

KurtStam:
…te as defaultValue

I fixed up the GitHub ITCase test too since it was failing for me.


KurtStam:
retest please

PR #621, opened 2017-09-20
chore: Updates deployent.json with latest Salesforce

zregvart:
**This won't pass tests until new version of connectors is released**

see https://github.com/syndesisio/connectors/pull/72 and https://github.com/syndesisio/connectors/pull/71

Now that Salesforce connectors have unified input/output DTO for actions
producing or consuming only ID java types in deployment.json need to be
updated.

Also adds to the `DeploymentDescriptorTest` checks that input/output
data shapes resemble those in the connector.


zregvart:
> Should we merge now or wait for the next connectors release ?

I would release the connectors first that way we can double check the `deployment.json`.

zregvart:
This should be OK to merge.

PR #625, opened 2017-09-21
fix(runtime): Add JBoss EA repo to generted pom.xml

rhuss:


jimmidyson:
Need to update test cases too.

rhuss:
@jimmidyson test should be good now.

PR #626, opened 2017-09-21
feat(setup): Add setup JAX-RS application for first time user setup

jimmidyson:
The setup application provides a single endpoint, at `/api/setup`. This is
a one time only handler: once it is updated, it is unusable. The endpoint
accepts the following methods:

`GET`: This checks if setup is required. It returns `204` (`No Content`) if
the setup handler is still enabled, and `410` (`Gone`) if it is already
configured and hence can't be called again.

`PUT`: This attempts to update the configuration as requested. This handler
will return `410` (`Gone`) if configuration is no longer required. If the
endpoint is still enabled, then the handler accepts a JSON body with the
following format:

```json
{
  "gitHubOAuthConfiguration": {
    "clientId": "yourclientid",
    "clientSecret": "topsecret"
  }
}
```

The format of this is extensible for any future first time user setup
steps.

/cc @kahboom @gashcrumb

### DIFF
--- a/github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
+++ b/github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
@@ -132,28 +132,34 @@ public class GitHubServiceITCase {
     @Test
     public void testCreateNewRepository() throws IOException, InterruptedException {
         String testRepo = REPO_NAME + "-create-new";
-        Repository repository = githubService.getRepository(testRepo);
-        int retryCount = 0;
-        while (repository != null) {
-            User apiUser = githubService.getApiUser();
-            client.delete("/repos/" + apiUser.getLogin() + "/" + testRepo);
-            this.wait(100l);
-            repository = githubService.getRepository(testRepo);
-            retryCount++;
-            if (retryCount > 50) {
-                break; //fail if the repo is still there after 5 seconds
-            }
-        }
-        Assert.assertNull(repository); // repository should not exist on GitHub
-
-        // Create Repository
-        repository = githubService.createRepository(testRepo);
-        Assertions.assertThat(repository).isNotNull();
-        Assertions.assertThat(repository.getName()).isEqualTo(testRepo);
-        System.out.println("Successfully created repository " + repository.getName());
-
         User apiUser = githubService.getApiUser();
-        client.delete("/repos/" + apiUser.getLogin() + "/" + testRepo);
+        try {
+            Repository repository = githubService.getRepository(testRepo);
+            int retryCount = 0;
+            if (repository != null) {
+                client.delete("/repos/" + apiUser.getLogin() + "/" + testRepo);
+                while (repository != null) {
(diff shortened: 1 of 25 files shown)


github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
```diff
@@ -132,28 +132,34 @@ public void testGetApiUser() throws IOException {
...
-        User apiUser = githubService.getApiUser();
-        client.delete("/repos/" + apiUser.getLogin() + "/" + testRepo);
+            // Create Repository
+            repository = githubService.createRepository(testRepo);
+            Assertions.assertThat(repository).isNotNull();
+            Assertions.assertThat(repository.getName()).isEqualTo(testRepo);
+            System.out.println("Successfully created repository " + repository.getName());
+        } catch (Exception e) {
+            e.printStackTrace();
+            Assert.fail(e.getMessage());
```
zregvart: perhaps just `throws Exception` in the method declaration, and we can remove this try-catch block.

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
```diff
@@ -132,28 +132,34 @@ public void testGetApiUser() throws IOException {
...
+            throw e;
+        } finally {
+            client.delete("/repos/" + apiUser.getLogin() + "/" + testRepo);
```
zregvart: Would this make more sense in a `@AfterClass` method?

jimmidyson:
@zregvart Thanks for review. Next time, please let me address comments in PR review cycle if they're small as they are.

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
jimmidyson: OK I'll swallow the exception in the finally block instead so we don't hide the real test failure reason.

github/src/test/java/io/syndesis/github/GitHubServiceITCase.java
jimmidyson: Hmm interesting idea. Or even an `@After` method I guess.

rhuss:
Would be a nice addition to purebot to e.g. respect e.g. a 'minor-change-requested' label. This could be set by the reviewer before approving and indicates for purebot to **not** merge even when reviewer approves. Only when this label is removed by the reviewee then its get merged. That way the reviewer does not need to come back to do the approval (which again adds to the review time).

I often also fall into the trap that I want to approve but have a minor nit-pick (but don't want to continue to monitor the PR)

jimmidyson:
@akieling As requested, you can use https://raw.githubusercontent.com/syndesisio/syndesis-rest/master/setup/reset-github-provider.sh to reset the github provider credentials on any target keycloak running on openshift. `reset-github-provider.sh <KEYCLOAK_URL>` is all you need to do.

PR #631, opened 2017-09-22
fix: sql stored connector id

zregvart:
SQL Stored connector id needs to be `sql-stored-connector` otherwise
when that ID is sent to verifier it will lookup the builtin `sql-stored`
Camel component.


zregvart:
retest this please

rhuss:
retest this please

PR #635, opened 2017-09-23
chore: upgrade connectors to 0.5.6

zregvart:
Also adds a bit of verbosity to DeploymentDescriptorTest to help
troubleshoot issues with fetching connectors.


project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
```diff
@@ -60,6 +60,7 @@
 
 @RunWith(Parameterized.class)
 public class DefaultProjectGeneratorTest {
+    private static final String CONNECTORS_VERSION = "0.5.6";
```
jimmidyson: Can we pass this as a system property to surefire rather than having to update each time we change the connector version?

zregvart: That was my first intention, but opted to make it as simple as possible, let me revisit this 👍

rhuss:
[merge]

PR #636, opened 2017-09-23
feat(runtime): Prefetch connector catalog

rhuss:
When creating project files, a `CamelCatalog` and `CamelConnectorCatalog`
is required to calculate the proper endpoint URL

Up to now, this is only done via Grape which does a live
introspection and download of connector jars.

A new plugin goal `syndesis:extract-connector-descriptor` has been
introduced to fetch this information offline.
By adding this to `META-INF/camel/camel-meta.json`, the runtime
can query this file and build up a CamelCatalog from this information only.
This avoids to reach out to the internet, which can have unexpected
side effects related to the performance.

-------

This is still work in progress. Actually the generation works nicely and has been introduced to the runtime pom.xml
During startup of the `ConnectorCatalog` this `/META-INF/camel/camel-meta.json` is parsed and the `CamelCatalog` as well as the `CamelConnectorCatalog` is build up.

It was tried to mimic the way how the `MavenArtifactProvider` resolves the required data, however it still fails.

When using the catalog on a prefetched connector, the following error happens:

```
2017-09-23 18:17:50.608 ERROR [-,,,] 1 --- [pool-5-thread-1] i.s.c.i.online.ActivateHandler           : Integration -KujxQDBjnvVpgWd0MgV : Failure

java.lang.IllegalArgumentException: Endpoint with scheme twitter-mention-connector has no syntax defined in the json schema
	at org.apache.camel.catalog.AbstractCamelCatalog.doAsEndpointUri(AbstractCamelCatalog.java:892) ~[camel-catalog-2.20.0.fuse-000101.jar!/:2.20.0.fuse-000101]
	at org.apache.camel.catalog.AbstractCamelCatalog.asEndpointUri(AbstractCamelCatalog.java:869) ~[camel-catalog-2.20.0.fuse-000101.jar!/:2.20.0.fuse-000101]
	at io.syndesis.connector.catalog.ConnectorCatalog.buildEndpointUri(ConnectorCatalog.java:164) ~[connector-catalog-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.visitor.EndpointStepVisitor.createEndpoint(EndpointStepVisitor.java:129) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.visitor.EndpointStepVisitor.createEndpoint(EndpointStepVisitor.java:125) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.visitor.EndpointStepVisitor.visit(EndpointStepVisitor.java:64) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.DefaultProjectGenerator.visitStep(DefaultProjectGenerator.java:219) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.DefaultProjectGenerator.lambda$generateFlowYaml$4(DefaultProjectGenerator.java:207) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at java.util.Optional.ifPresent(Optional.java:159) ~[na:1.8.0_131]
	at io.syndesis.project.converter.DefaultProjectGenerator.generateFlowYaml(DefaultProjectGenerator.java:184) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.project.converter.DefaultProjectGenerator.generate(DefaultProjectGenerator.java:157) ~[project-generator-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.controllers.integration.online.ActivateHandler.createProjectFiles(ActivateHandler.java:251) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.controllers.integration.online.ActivateHandler.execute(ActivateHandler.java:123) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.controllers.integration.IntegrationController.lambda$callStatusChangeHandler$10(IntegrationController.java:175) [controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_131]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[na:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) ~[na:1.8.0_131]
	at java.lang.Thread.run(Thread.java:748) ~[na:1.8.0_131]
```

The following connectors have been preloaded from the plugin's dependencies:

```
2017-09-23 18:15:32.127  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for org.apache.camel:camel-salesforce:2.20.0.fuse-000101
2017-09-23 18:15:32.128  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-on-update-connector:0.5.5
2017-09-23 18:15:32.130  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-on-delete-connector:0.5.5
2017-09-23 18:15:32.135  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-update-sobject-connector:0.5.5
2017-09-23 18:15:32.143  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-upsert-sobject-connector:0.5.5
2017-09-23 18:15:32.156  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-create-sobject-connector:0.5.5
2017-09-23 18:15:32.160  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-get-sobject-connector:0.5.5
2017-09-23 18:15:32.163  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-get-sobject-with-id-connector:0.5.5
2017-09-23 18:15:32.166  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-delete-sobject-connector:0.5.5
2017-09-23 18:15:32.167  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:salesforce-delete-sobject-with-id-connector:0.5.5
2017-09-23 18:15:32.170  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for io.syndesis:sql-stored-connector:0.5.5
2017-09-23 18:15:32.172  INFO [-,,,] 1 --- [           main] i.s.connector.catalog.ConnectorCatalog   : Prefetched meta for org.apache.camel:camel-sql:2.20.0.fuse-000101
```

Don't think its a major issue, but happy for any hint to fix this as I'm on the road tomorrow and this is still 
considered to be a blocker for TP1. 'will continue tomorrow, but who will fix this before me, will get (a) one pint of franconian beer and (b) a mini-jar of fresh Jolokia pepper powder from 2017 ;-)

@lburgazzoli @jimmidyson @zregvart @davsclaus is this good enough of an offer ;-) ?

### DIFF
--- a/connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
+++ b/connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2016 Red Hat, Inc.
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -15,7 +15,21 @@
  */
 package io.syndesis.connector.catalog;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.camel.catalog.CamelCatalog;
+import org.apache.camel.catalog.CollectionStringBuffer;
 import org.apache.camel.catalog.DefaultCamelCatalog;
 import org.apache.camel.catalog.connector.CamelConnectorCatalog;
 import org.apache.camel.catalog.connector.DefaultCamelConnectorCatalog;
@@ -24,21 +38,23 @@ import org.apache.camel.catalog.maven.MavenArtifactProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.net.URISyntaxException;
-import java.util.Map;
-
 public class ConnectorCatalog {
 
     private static final Logger LOG = LoggerFactory.getLogger(ConnectorCatalog.class);
+    private static final String CONNECTOR_SCHEMAS = "/META-INF/camel/camel-meta.json";
 
     private final CamelConnectorCatalog connectorCatalog;
     private final CamelCatalog camelCatalog;
     private final MavenArtifactProvider maven;
 
+    private final Set<String> prefetchedConnectors = new HashSet<>();
+
     public ConnectorCatalog(ConnectorCatalogProperties props) {
 
         connectorCatalog = new DefaultCamelConnectorCatalog();
         camelCatalog = new DefaultCamelCatalog(true);
+
+        prefetchConnectors();
 
         maven = new DefaultMavenArtifactProvider();
 
@@ -51,7 +67,124 @@ public class ConnectorCatalog {
         }
     }
 
+    private void prefetchConnectors() {
+        try (InputStream is = getClass().getResourceAsStream(CONNECTOR_SCHEMAS)) {
+            if (is != null) {
+                ArrayNode descriptors = (ArrayNode) new ObjectMapper().readTree(is);
+                for (JsonNode descriptor : descriptors) {
+                    JsonNode gav = descriptor.get("gav");
+                    if (gav == null) {
+                        continue;
+                    }
+                    String gavCoords[] = gav.textValue().split(":");
+                    if (gavCoords.length < 3) {
+                        continue;
+                    }
+                    LOG.info("Prefetched meta for {}:{}:{}", gavCoords[0], gavCoords[1], gavCoords[2]);
+                    prefetchedConnectors.add(gav.textValue());
+
+                    JsonNode connector = descriptor.get("connector");
+                    ObjectNode component = (ObjectNode) descriptor.get("component");
+                    if (connector != null) {
+                        addConnector(gavCoords, connector, component);
+                    }
+                    if (component != null) {
+                        addComponent(component);
+                    }
+                }
+            }
+        } catch (IOException exp) {
+            LOG.error("Cannot initialize connectors from {} : {}", CONNECTOR_SCHEMAS, exp.getMessage(), exp);
+        }
+    }
+
+    private void addComponent(ObjectNode component) {
+        ObjectNode meta = (ObjectNode) component.get("meta");
+        Iterator<Map.Entry<String, JsonNode>> it = meta.fields();
+        while (it.hasNext()) {
+            Map.Entry<String, JsonNode> entry = it.next();
+            String json = entry.getValue().asText();
+            String javaType = extractJavaType(json);
+            if (javaType != null) {
+                camelCatalog.addComponent(entry.getKey(), javaType, json);
+            }
+        }
+    }
+
+    private void addConnector(String[] gavCoords, JsonNode connector, ObjectNode component) {
+        JsonNode meta = connector.get("meta");
+        connectorCatalog.addConnector(
+            gavCoords[0], gavCoords[1], gavCoords[2],
+            meta.get("name").textValue(),
+            meta.get("scheme").textValue(),
+            meta.get("javaType").textValue(),
+            meta.get("description").textValue(),
+            extractLabels(meta),
+            extractJson(connector, "meta"),
+            extractJsonTextString(connector, "schema"),
+            extractJsonTextString(component, "schema"));
+    }
+
+    private String extractJavaType(String json) {
+        try {
+            ObjectNode node = (ObjectNode) new ObjectMapper().readTree(json);
+            JsonNode c = node.get("component");
+            if (c != null) {
+                JsonNode ret = c.get("javaType");
+                if (ret != null) {
+                    return ret.asText();
+                }
+            }
+            return null;
+        } catch (IOException e) {
+            return null;
+        }
+    }
+
+    private String extractJsonTextString(JsonNode descriptor, String key) {
+        if (descriptor != null && descriptor.hasNonNull(key)) {
+            // TODO: If this is a real JSON object, print it as JSON
+            // I.e. when JsonSchemaHelper is 'fixed', and
+            // ExtractConnectorDescriptorsMojo updated to store real json objects for
+            // JSONNodes and not only text
+            return descriptor.get(key).asText();
+        } else {
+            return null;
+        }
+    }
+
+    private String extractJson(JsonNode descriptor, String key) {
+        if (descriptor != null && descriptor.hasNonNull(key)) {
+            return getJsonAsString(descriptor.get(key));
+        } else {
+            return null;
+        }
+    }
+
+    private String getJsonAsString(JsonNode jsonNode) {
+        try {
+            return new ObjectMapper().writeValueAsString(jsonNode);
+        } catch (JsonProcessingException e) {
+            return null;
+        }
+    }
+
+    private String extractLabels(JsonNode tree) {
+        Iterator<JsonNode> it = tree.withArray("labels").iterator();
+
+        CollectionStringBuffer csb = new CollectionStringBuffer(",");
+        while (it.hasNext()) {
+            String text = it.next().textValue();
+            csb.append(text);
+        }
+        return csb.toString();
+    }
+
     public final void addConnector(String gav) {
+        if (prefetchedConnectors.contains(gav)) {
+            return;
+        }
+
         String[] splitGAV = gav.split(":");
         if (splitGAV.length == 3) {
             String groupId = splitGAV[0];
@@ -67,11 +200,11 @@ public class ConnectorCatalog {
     public String buildEndpointUri(String scheme, Map<String, String> options) throws URISyntaxException {
         String result = camelCatalog.asEndpointUri(scheme, options, false);
         // we need to strip off the colon bit.
-        if( result.equals(scheme+":")) {
+        if (result.equals(scheme + ":")) {
             result = scheme;
         }
-        if( result.startsWith(scheme+":?")) {
-            result = scheme+result.substring(scheme.length()+1);
+        if (result.startsWith(scheme + ":?")) {
+            result = scheme + result.substring(scheme.length() + 1);
         }
         return result;
     }
--- a/syndesis-maven-plugin/pom.xml
+++ b/syndesis-maven-plugin/pom.xml
@@ -38,11 +38,75 @@
     <dependency>
       <groupId>io.syndesis</groupId>
       <artifactId>model</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-context</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-web</artifactId>
+        </exclusion>
+        <exclusion>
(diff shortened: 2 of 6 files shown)


rhuss:
The issue here is that [JSonSchemaHelper.parseJsonSchema()](https://github.com/apache/camel/blob/master/camel-core/src/main/java/org/apache/camel/runtimecatalog/JSonSchemaHelper.java#L44-L123) **can only parse pretty-printed JSON with a very special format**, but I put it on a single line (as I parse it first and then serialize it back as I'm working here with JsonNode objects.

Come on guys, this can't be true. If this can't be changed (should be easy though, happy to send a PR), then please don't call this JsonSchema. Call it CamelSchema or whatever. But not JsonSchema and not Json.

'wonder really why nobody already stumbled upon this. // @davsclaus @zregvart

davsclaus:
@rhuss for the 110 time there is a CAMEL ticket about it. And we do not want to have a runtime dependency on a heavy json parser like jackson that causes all sorts of classpath issues when other libraries uses different version of it etc.

Recently a json spec was passed which will make it easier to plugin and use a lighter library in the future, or even shade a library into camel-core if necessary.

davsclaus:
And btw @rhuss it seems the bot wants the beer it reported a gazillion issues.

rhuss:
@davsclaus thx

davsclaus:
The CAMEL ticket is: https://issues.apache.org/jira/browse/CAMEL-10969 and https://issues.apache.org/jira/browse/CAMEL-11005

rhuss:
> And btw @rhuss it seems the bot wants the beer it reported a gazillion issues.

good idea, will fix it, @davsclaus . Then it's not so evident anymore that I reused quite a bit from the camel code ;-P (its not so easy to build up a catalog programmatially, but happy for any pointer how this can be done more easily)

davsclaus:
@rhuss okay working on using a better parser, json-simple is maintained in v2 which we can use.

There is a branch here
https://github.com/apache/camel/tree/json-simple

And you can turn on|off prettyPrint mode on the camel-connector-maven-plugin (default on)
https://github.com/apache/camel/blob/json-simple/connectors/camel-connector-maven-plugin/src/main/java/org/apache/camel/maven/connector/ConnectorMojo.java#L81

And you have json in pretty print such as (its hardcoded to use tab in json-simple so we cannot change to use 2 spaces etc)
https://github.com/apache/camel/blob/json-simple/connectors/examples/foo-connector/src/main/resources/camel-connector-schema.json

The API is the same as before, you just use JSonSchemaHelper from camel-catalog
https://github.com/apache/camel/blob/json-simple/platforms/camel-catalog/src/main/java/org/apache/camel/catalog/JSonSchemaHelper.java#L42

... which now supports:
- old single-line mode
- pretty printed mode
- or any json mode (assuming the parser works well)

Mind that outputting in pretty print mode, may affect older tooling such as Fuse IDE that may use an older version of camel-catalog with its parser. But if they upgrade to use Camel 2.20.0 then they should be okay.

We are running tests on the branch to see if it causes and problems, if not then its very likely to be included in Camel 2.20.0 as json-simple is shaded into camel-core and camel-catalog and its a single JAR so it does not drag in a lot of complexity/crap or potential problems with using jackson etc.

davsclaus:
@rhuss @lburgazzoli @jimmidyson @zregvart @gashcrumb okay so the master branch of Apache Camel now comes with a better json parser embedded (based on simple-json v2) which supports json files as before but also in pretty print mode (eg real json schema). The API from camel-catalog and camel-core is 100% the same. 

The camel-catalog-maven-plugin now outputs the generated json schema it spits out in pretty print mode, but you can turn it off to make it as it was before. 

And the camel-connector component can then now read any kind of json schema file you throw at it.

The ticket at Apache Camel about this is: https://issues.apache.org/jira/browse/CAMEL-10969

rhuss:
@davsclaus A.w.e.s.o.m.e !! Thanks a lot, that make things so much easier and more robust. I'm going to revert back to the original variation of this PR when the camel upstream change bubbled up to a pre-release version.

rhuss:
retest this please

rhuss:
retest this please

rhuss:
@zregvart @jimmidyson @KurtStam @iocanel Time for a review ? This PR is ready for prime time and helps in creating project files (and avoids external download for creating the CamelCatalog)

connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
```diff
@@ -1,12 +1,12 @@
...
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
zregvart: this might break the license check (wonder why it did not?)

connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
```diff
@@ -51,7 +67,124 @@ public ConnectorCatalog(ConnectorCatalogProperties props) {
         }
     }
 
+    private void prefetchConnectors() {
+        try (InputStream is = getClass().getResourceAsStream(CONNECTOR_SCHEMAS)) {
+            if (is != null) {
+                ArrayNode descriptors = (ArrayNode) new ObjectMapper().readTree(is);
+                for (JsonNode descriptor : descriptors) {
+                    JsonNode gav = descriptor.get("gav");
+                    if (gav == null) {
```
zregvart: don't we control what's in `/META-INF/camel/camel-meta.json` when would `gav` be `null`?

syndesis-maven-plugin/pom.xml
```diff
@@ -38,11 +38,75 @@
     <dependency>
       <groupId>io.syndesis</groupId>
       <artifactId>model</artifactId>
+      <exclusions>
+        <exclusion>
```
zregvart: when I think about this perhaps:

```xml
<exclusion>
  <groupId>*</groupId>
  <artifactId>*</artifactId>
</exclusion>
```

might be better (I wish I did that first time around)

syndesis-maven-plugin/pom.xml
```diff
@@ -38,11 +38,75 @@
...
+          <artifactId>javax.ws.rs-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
       <groupId>io.syndesis</groupId>
       <artifactId>dao</artifactId>
+      <exclusions>
+        <exclusion>
```
zregvart: same thing as above

syndesis-maven-plugin/pom.xml
```diff
@@ -120,6 +213,100 @@
       <groupId>io.syndesis</groupId>
       <artifactId>rest</artifactId>
       <scope>runtime</scope>
+      <exclusions>
+        <exclusion>
```
zregvart: I think `*` exclusion would be better here also

syndesis-maven-plugin/src/main/java/io/syndesis/maven/ExtractConnectorDescriptorsMojo.java
```diff
@@ -0,0 +1,284 @@
...
+        try {
+            PluginDescriptor desc = (PluginDescriptor) getPluginContext().get("pluginDescriptor");
+            List<Artifact> artifacts = desc.getArtifacts();
+            ProjectBuildingRequest buildingRequest =
+                new DefaultProjectBuildingRequest(session.getProjectBuildingRequest());
+            buildingRequest.setRemoteRepositories(remoteRepositories);
+            for (Artifact artifact : artifacts) {
+                ArtifactResult result = artifactResolver.resolveArtifact(buildingRequest, artifact);
+                File jar = result.getArtifact().getFile();
+                classLoader = createClassLoader(jar);
```
zregvart: perhaps try-with-resources for the `classLoader`?

syndesis-maven-plugin/src/main/java/io/syndesis/maven/ExtractConnectorDescriptorsMojo.java
```diff
@@ -0,0 +1,284 @@
...
+                try {
+                    classLoader.close();
+                } catch (IOException ignored) {
+
+                }
+            }
+        }
+    }
+
+    private URLClassLoader createClassLoader(File jar) throws MalformedURLException {
```
zregvart: I wonder if we need a classloader or can use just `JarInputStream`, we're just loading JSON/properties files not classes right?

connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
rhuss: Seem like that markup doesn't harm for this plugin. I regenerated the header nevertheless.

connector-catalog/src/main/java/io/syndesis/connector/catalog/ConnectorCatalog.java
rhuss: Yes, we do. So maybe better throw an exception then except ignoring it?

syndesis-maven-plugin/pom.xml
rhuss: Ok, changed this. But not sure why the exclusions are required anyway?

syndesis-maven-plugin/src/main/java/io/syndesis/maven/ExtractConnectorDescriptorsMojo.java
rhuss: I did try in the beginning but got ditched by the code checks (don't remember why anymore, but that was the quickest solution)

syndesis-maven-plugin/src/main/java/io/syndesis/maven/ExtractConnectorDescriptorsMojo.java
rhuss: Yes, just tried to use a JarInputStream. But reading real data from a JarInputStream is quite tedious (without proper helper which I couldn't find any). So would like to keep that. I like the simplicity of getResourceAsStream.

syndesis-maven-plugin/pom.xml
rhuss: Tried that, but are running in tons of issues: Maven duplicate plugin croaks a lot plus when running it:

```
[ERROR] Failed to execute goal io.syndesis:syndesis-maven-plugin:0.1-SNAPSHOT:generate-mapper-inspections (generate-mapper-inspections) on project runtime: Execution generate-mapper-inspections of goal io.syndesis:syndesis-maven-plugin:0.1-SNAPSHOT:generate-mapper-inspections failed: A required class was missing while executing io.syndesis:syndesis-maven-plugin:0.1-SNAPSHOT:generate-mapper-inspections: io/syndesis/core/Json
```

rhuss: So reverted it again (sorry, have to prioritize ...)

rhuss:
@zregvart are we good to merge ? I tried to adressed some of your issue, but struggled. Let's do this either later or ...

PR #641, opened 2017-09-25
feat(openshift): Go direct to OpenShift API rather than using token replacing proxy

jimmidyson:


jimmidyson:
retest this please

PR #642, opened 2017-09-25
fix: if the status of the integration revision ...

### LINKED ISSUE #638: Huge number of Integration revisions rows in the database
On staging environment we have two integrations with 434046 and 332934 rows containing integration revisions. Most likely we're only adding rows to revisions and not replacing them. Could explain the out of memory errors we have seen also.

zregvart:
...does not change update the last revision

Currently we add a new integration revision for every status check,
about one every minute, this leads to a huge number of rows being added
to the database.

This change updates the last integration revision if the status of the
last revision does not change with the update. This way we have the
number of revisions tied to the number of state changes.

Fixes #638

### DIFF
--- a/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
+++ b/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
@@ -197,7 +197,7 @@ public class IntegrationController {
                         .withCurrentState(IntegrationRevisionState.from(update.getStatus()));
 
                     //replace revision
-                    integration.getDeployedRevision().map(revisions::remove);
+                    revisions.remove(revision);
 
                     final IntegrationRevision last = integration.lastRevision();
                     if (IntegrationRevisionState.from(update.getStatus()).equals(last.getCurrentState())) {
--- a/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
+++ b/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
@@ -55,8 +55,8 @@ public class IntegrationController {
     private final EventBus eventBus;
     private final ConcurrentHashMap<Integration.Status, StatusChangeHandlerProvider.StatusChangeHandler> handlers = new ConcurrentHashMap<>();
     private final Set<String> scheduledChecks = new HashSet<>();
-    private ExecutorService executor;
-    private ScheduledExecutorService scheduler;
+    /* default */ ExecutorService executor;
+    /* default */ ScheduledExecutorService scheduler;
 
     private static final long SCHEDULE_INTERVAL_IN_SECONDS = 60;
 
@@ -159,7 +159,7 @@ public class IntegrationController {
         return "Integration " + integration.getId().orElse("[none]");
     }
 
-    private void callStatusChangeHandler(StatusChangeHandlerProvider.StatusChangeHandler handler, String integrationId) {
+    /* default */ void callStatusChangeHandler(StatusChangeHandlerProvider.StatusChangeHandler handler, String integrationId) {
         executor.submit(() -> {
             Integration integration = dataManager.fetch(Integration.class, integrationId);
             String checkKey = getIntegrationMarkerKey(integration);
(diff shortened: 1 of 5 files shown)


jimmidyson:
Why is status even included in revision? Shouldn't we just be storing history of the desired spec without status?

zregvart:
> Why is status even included in revision? Shouldn't we just be storing history of the desired spec without status?

As I read the intent of the [design proposal](https://github.com/syndesisio/syndesis-project/blob/master/proposals/044-integration-model.md) is to remove the state from the Integration and use only the state from the last revision.

jimmidyson:
Yes thanks for the pointer. It doesn't make much sense to me. Reverting to a previous revision wouldn't/shouldn't include the state in that so I don't know why we keep the state in the revision?

Thoughts @iocanel @rhuss?

zregvart:
I guess builds pass based on what proxy they hit for JBoss EA repository, one that has our dependencies or one that has not.

rhuss:
> Yes thanks for the pointer. It doesn't make much sense to me. Reverting to a previous revision wouldn't/shouldn't include the state in that so I don't know why we keep the state in the revision?

The reason is that we have revisions with different states which are important for an Integration:

* 0 or 1 revision in state _Draft_ which is the one currently worked on and which is not deployed yet
* 0 or 1 revision in state _Deployed_ this is the active revision.
* 0 ... n revisions in state _Undeployed_ which is the history.

Not sure how this should be modelled with a single state on the Integration only.

rhuss:
I we would modell the _draft_ use case differently (e.g. having a non-revision kind of object describing the currently worked on integration which is converted to a revision when deployed), then of course we could go with keeping the current and target state on the integration itself. Any ideas here, @jimmidyson ?

On the other end it might be interesting to keep the information whether a revision was in error state (but not 100% sure about how errors state is to be modelled)

rhuss:
BTW, this issue could be also the reason why we run out of memory with syndesis-rest on staging, because the infinispan cache blew up ?

controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
```diff
@@ -197,10 +197,17 @@ private void callStatusChangeHandler(StatusChangeHandlerProvider.StatusChangeHan
                         .withCurrentState(IntegrationRevisionState.from(update.getStatus()));
 
                     //replace revision
-                    revisions.remove(revision);
+                    integration.getDeployedRevision().map(revisions::remove);
```
iocanel: We already have a reference of `revision` and `revsisions` what is the purpose of re-obtaining the revision (given that we only compare for equality the revisionId and ignore the rest of the object)?

jimmidyson:
I do feel quite strongly that we shouldn't be storing state in the revision history. If we think about the integration history page in the UI, then the user would see multiple revisions for an identical integration spec to revert to, only difference being the state of that revision. That doesn't seem to work at all.

So it looks like we only care about the draft revision, which I would model differently as @rhuss stated. As revisions are immutable anyway, having the draft state, which I would consider a naturally mutable resource, adding that to the revision history as an immutable resource doesn't make sense, unless any draft revision is replaced in the revision history I guess? Doesn't feel right though.

chirino:
I agree.. current integration state should not be in the history.

rhuss:
I'm fine with @jimmidyson suggestion. Question is whether this should go into TP1 or can we work around this for now ?

zregvart:
I think we first need to address the jsondb rown churn as this is most pressing and is blocking the release. Any design changes can be revisited subsequently, preferably with a discussion and a PR on the design proposal.

jimmidyson:
Yeah I agree with @zregvart, let's get this merged ASAP and then figure out how to migrate to better designed model.

jimmidyson:
retest this please

zregvart:
I would like to address the review comment by @iocanel also.

jimmidyson:
Is this in a final state to review now?

zregvart:
> Is this in a final state to review now?

Should be, thanks 🥇

PR #645, opened 2017-09-27
refactor(metadata): update metadata response from verifier

zregvart:
This updates the verifier's response for metadata as implemented in
syndesisio/syndesis-verifier#40


KurtStam:
retest this please

PR #649, opened 2017-09-28
fix(grammar): proper pronoun use

zregvart:
Fixes: syndesisio/syndesis-ui#1031


dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -375,7 +375,7 @@
         },{
           "id": "io.syndesis:salesforce-get-sobject-connector:latest",
           "name": "Fetch record",
-          "description": "Fetches a record from Salesforce by it's identifier value.",
+          "description": "Fetches a record from Salesforce by its identifier value.",
```
jimmidyson: :nerd_face:

Issue #655, opened 2017-09-30
Error 500 on /api/setup endpoint


akieling:
**Current behavior**
Invoking /api/setup is throwing error with status code 500

**Expected behavior**
It should not throw any errors

**Steps to reproduce**
1. Download https://github.com/kahboom/syndesis-ui/tree/952-github-oauth
1. Start the app
1. Go to https://localhost:4200/setup
1. On step 2, enter client ID and client secret and press Connect GitHub button (this action sends a PUT to `/api/setup` as specified in https://github.com/syndesisio/syndesis-rest/pull/626)
1. Check the browser console and the syndesis-rest logs

**Logs**
`2017-09-29 18:30:52.542 ERROR [-,,,] 1 --- [  XNIO-3 task-6] io.undertow.request                      : UT005023: Exception handling request to /api/setup

java.lang.NullPointerException: null
at io.syndesis.runtime.SyndesisCorsConfiguration.lambda$corsFilter$0(SyndesisCorsConfiguration.java:46) ~[classes!/:0.1-SNAPSHOT]
at org.springframework.web.filter.CorsFilter.doFilterInternal(CorsFilter.java:87) ~[spring-web-4.3.10.RELEASE.jar!/:4.3.10.RELEASE]
at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:107) ~[spring-web-4.3.10.RELEASE.jar!/:4.3.10.RELEASE]
at io.undertow.servlet.core.ManagedFilter.doFilter(ManagedFilter.java:61) ~[undertow-servlet-1.4.18.Final.jar!/:1.4.18.Final]
at io.undertow.servlet.handlers.FilterHandler$FilterChainImpl.doFilter(FilterHandler.java:131) ~[undertow-servlet-1.4.18.Final.jar!/:1.4.18.Final]`

rhuss:
@akieling I just pushed a fix for this very NPE to GitHub, the image should be build soon after and staging should be updated then in around 30mins.

Please let me know whether this helps. I tried to test the setup locally but failed. Will check tomorrow again.

akieling:
@rhuss that seems to fix it but now I'm getting another error. Maybe an OpenShift route configuration?
```
Error: Response with status: 404 Not Found for URL: https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/api/setup
```

rhuss:
[...]

PR #656, opened 2017-09-30
fix: NPE in Cors filter when no path is set

rhuss:


rhuss:
Fixes hopefully #655

rhuss:
self-merge as it is urgent (and possibly to small of  a change to fail).

PR #665, opened 2017-10-02
refactor(integration): Set builder memory to 300m

rhuss:
Ideally this should be configurable as well as the requests/limits of the builder pod.

For our cluster we have 512 MB limit so a mx of 500M is too large.
300M should be safe, however, as said, this should be configurable.


rhuss:
@zregvart quick review if possible ?

rhuss:
retest this please

PR #667, opened 2017-10-03
specific cache key based on poms

KurtStam:
You can only write to the cache once, so the key needs to update if dependencies change, or else you keep missing the and re downloading the those new dependencies. Taking a hash of all poms in the project seems like a pretty good key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ jobs:
       - restore_cache:
           key: syndesis-rest-m2-{{ checksum "pom.xml" }}-{{ checksum "connector-catalog/pom.xml" }}-{{ checksum "controllers/pom.xml" }}-{{ checksum "core/pom.xml" }}-{{ checksum "credential/pom.xml" }}-{{ checksum "dao/pom.xml" }}-{{ checksum "github/pom.xml" }}-{{ checksum "jsondb/pom.xml" }}-{{ checksum "model/pom.xml" }}-{{ checksum "model2/pom.xml" }}-{{ checksum "openshift/pom.xml" }}-{{ checksum "project-generator/pom.xml" }}-{{ checksum "rest/pom.xml" }}-{{ checksum "syndesis-maven-plugin/pom.xml" }}-{{ checksum "runtime/pom.xml" }}-{{ checksum "verifier/pom.xml" }}-{{ checksum "inspector/pom.xml" }}-{{ checksum "setup/pom.xml" }}
           key: syndesis-rest-m2-{{ checksum "pom.xml" }}
+          key: syndesis-rest-m2
 
       - run:
           name: Maven
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,14 +30,16 @@ jobs:
       - checkout
 
       - restore_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -30,14 +30,16 @@ jobs:
       - checkout
 
       - restore_cache:
-          key: syndesis-rest-m2
+          key: syndesis-rest-m2-{{ checksum "pom.xml" }}-{{ checksum "connector-catalog/pom.xml" }}-{{ checksum "controllers/pom.xml" }}-{{ checksum "core/pom.xml" }}-{{ checksum "credential/pom.xml" }}-{{ checksum "dao/pom.xml" }}-{{ checksum "github/pom.xml" }}-{{ checksum "jsondb/pom.xml" }}-{{ checksum "model/pom.xml" }}-{{ checksum "model2/pom.xml" }}-{{ checksum "openshift/pom.xml" }}-{{ checksum "project-generator/pom.xml" }}-{{ checksum "rest/pom.xml" }}-{{ checksum "syndesis-maven-plugin/pom.xml" }}-{{ checksum "runtime/pom.xml" }}-{{ checksum "verifier/pom.xml" }}-{{ checksum "inspector/pom.xml" }}-{{ checksum "setup/pom.xml" }}
+          key: syndesis-rest-m2-{{ checksum "pom.xml" }}
```
rhuss: Double key ?

KurtStam: Yes that's a fallback.

KurtStam:
@rhuss If it's guaranteed the top level pom changes with all changes to plugin and build dependencies in sub module poms then that should work. Are you sure? Remember you can save the cache only once.

rhuss:
@KurtStam well, it should be. It's as fragile as listing all modules here when e.g. a module appears or disappears (like soon the github module). Nobody will remember to update this file.

jimmidyson:
This change will actually slow down next builds when any POM changes as all dependencies will need to be re-downloaded (cache invalidated). Is that what we want?

The problem with the current key is that it only causes a growth in the cache, but at least existing deps don't need to be redownloaded.

KurtStam:
@jimmidyson the issue is that you can only write to the cache once. So it will never add new dependencies, and it will keep on downloading them with every build. We can use multiple keys as fallbacks so we don't need to re-download from scratch, but the top level one should be as specific as possible, that way the next build is fully cached.

jimmidyson:
Ah I missed `Cache is immutable on write` from https://circleci.com/docs/2.0/caching

KurtStam:
@jimmidyson  exactly

KurtStam:
@jimmidyson @rhuss Can this be merged?

.circleci/config.yml
jimmidyson: I think we should add a `syndesis-rest-m2` as a final fallback key just for fun.

KurtStam:
@jimmidyson fallback added

KurtStam:
hmm this change is in a circle-ci config file. I guess our build it broken in general?

KurtStam:
@jimmidyson can you approve one more time, now that the build is fixed?

PR #675, opened 2017-10-04
refactor: Use a reference to an existing imagestream tag for the s2i builder

rhuss:
It is assumed that this stream has been created outside alread, e.g. by the templating mechanism.

This PR requires an update of the templates (PR pending)


rhuss:
retest this please

zregvart:
I'm not exactly sure about the issue this is fixing, so I could be wrong: should there be a corresponding `application.yml` change in here also?

rhuss:
The issue is, that we are creating the image stream for the build image on our own on the fly. However, this image stream should be created outside of syndesis-rest as it looks different for different environments.

There is a sane default for the image stream tag name 's2i-java:2.0' which is created by openshift-templates now (pointing to the docker hub hosted s2i builder image). For production on the ignite cluster, we will have a dedicated builder stream in another project. The image stream tag name can be configured and is set in the corresponding openshift-template PR (see above for the link)

PR #677, opened 2017-10-05
refactor(project-generator): remove camel-test dependency

zregvart:
This removes `camel-test` dependacy from the generated integration POM.
Should be few bytes less to transfer in the build.


rhuss:
retest this please

PR #678, opened 2017-10-06
fix: Proper build and deployment handling

rhuss:
For now we are triggering a deployment explicitly, to circumvent
issue on our target platform which has problems with deployment triggers.

However I was not able to disable image triggers on the deployment, otherwise
imagestream could not be used. This cause a double deployment on
platforms with working triggers
This is not nice but should not harm to much.

### DIFF
--- a/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
+++ b/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
@@ -360,11 +360,12 @@ public class ActivateHandler extends BaseHandler implements StatusChangeHandlerP
 
     private class BuildStepPerformer {
         private final List<String> stepsPerformed;
-        private Integration integration;
+        private final Integration integration;
 
         BuildStepPerformer(Integration integration) {
             this.integration = integration;
-            this.stepsPerformed = integration.getStepsDone().orElseGet(ArrayList::new);
+            this.stepsPerformed = integration.getStepsDone().isPresent() ?
+                new ArrayList<>(integration.getStepsDone().get()) : new ArrayList<>();
         }
 
         /* default */ void perform(String step, IoCheckedFunction<Integration> callable) throws IOException {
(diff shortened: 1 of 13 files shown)


controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
```diff
@@ -318,4 +350,36 @@ private static boolean hasComponentProperties(Map<String, String> properties, Wi
...
+         void apply(T t) throws IOException;
+    }
+
+    private class BuildStepPerformer {
+        private final List<String> stepsPerformed;
+        private Integration integration;
+
+        BuildStepPerformer(Integration integration) {
+            this.integration = integration;
+            this.stepsPerformed = integration.getStepsDone().orElseGet(ArrayList::new);
```
zregvart: This list will be unmodifiable if coming from `integration`, so stepsPerformed.add below might fail.

zregvart:
Checkstyle complains with:

```
[INFO] There is 1 error reported by Checkstyle 6.19 with checkstyle.xml ruleset.

[ERROR] src/main/java/io/syndesis/controllers/integration/StatusChangeHandlerProvider.java:[35,19] (modifier) ModifierOrder: 'private' modifier out of order with the JLS suggestions.
```

openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
```diff
@@ -178,7 +200,7 @@ private void ensureBuildConfig(String name, DeploymentData deploymentData, Strin
               .withNewSourceStrategy()
                 .withNewFrom().withKind("ImageStreamTag").withName(builderStreamTag).endFrom()
                 .withIncremental(true)
-                .withEnv(new EnvVar("MAVEN_OPTS","-XX:+UseG1GC -XX:+UseStringDeduplication -Xmx500m", null))
+                .withEnv(new EnvVar("MAVEN_OPTS","-XX:+UseG1GC -XX:+UseStringDeduplication -Xmx300m", null))
```
zregvart: We might want to externalize this at some point...

rhuss: yeah, I even has comment there (but got lost ;) Happy for every PR

controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
rhuss: good catch, hadn't yet a situation where one of the steps failed. I hope to get rid of keeping such kind of state on the integration in the near future, though.

PR #686, opened 2017-10-08
fix: Missing namespace for builder imagestream

rhuss:
Also rearranged order for creation of resources a bit to make it more robuts


rhuss:
Self merging, as probably no one is available on Sunday ? (but need this PR ASAP)

rhuss:
[merge]

PR #688, opened 2017-10-09
Adding a small SB app which generates an integration project from a CLI.  Used to prime a builder image.

chirino:


syndesis-builder-image-generator/src/main/java/io/syndesis/image/Application.java
```diff
@@ -0,0 +1,184 @@
...
+                            public io.syndesis.integration.model.steps.Step visit(StepVisitorContext stepContext) {
+                                return null;
+                            }
+                        };
+                    }
+                }
+            )
+        );
+
+        try (InputStream is = new DefaultProjectGenerator(new ConnectorCatalog(catalogProperties), generatorProperties, registry).generate(request)) {
```
rhuss: Wouldn't it be easier to just create the pom.xml with `ProjectGenerator.generatePom()` and only use this pom.xml to download all deps ?

rhuss: you wouldn't need to deal with the tar archive then, but can directly write out this single file. We could even create simple pom.xml template on our own and write out the action gav's as deps directly so we don't have to create a dummy integration, which feels a bit hacky.

rhuss: See https://github.com/rhuss/ipaas-rest/blob/41b954a25817498ab68cb4351013ffc7d0b56bfe/project-generator/src/main/java/io/syndesis/project/converter/DefaultProjectGenerator.java#L217-L227 how the actions are extracted and https://github.com/rhuss/ipaas-rest/blob/fc03a6e4dbfd5d73760ffb4c40d30e52e272d1d3/project-generator/src/main/resources/io/syndesis/project/converter/templates/pom.xml.mustache#L74-L80 how its interpolated as dependencies.

pom.xml
```diff
@@ -197,6 +197,7 @@
     <module>verifier</module>
     <module>inspector</module>
     <module>setup</module>
+    <module>syndesis-builder-image-generator</module>
```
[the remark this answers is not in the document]
chirino: So this impact artifact names, the jar files.  Not prefixing with syndesis makes it much less obvious when just looking a jars where they come from.

syndesis-builder-image-generator/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+
+  <build>
+    <finalName>syndesis-builder-image-generator</finalName>
+
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+      </resource>
+      <resource>
+        <directory>../runtime/src/main/resources</directory>
```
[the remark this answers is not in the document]
chirino: But it's much more simple on the build system.

syndesis-builder-image-generator/src/main/java/io/syndesis/image/Application.java
chirino: I could do, but I'm not sure the project would properly build via 'mvn package'

chirino:
@rhuss the generate method is now much more simplified as it bypasses the tar bits.
I think generating the pom from an integration is ok for now, we can always bypass this too in the future if it becomes an issue.

pom.xml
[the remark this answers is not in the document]
rhuss: agreed, but I think also that consistency is more important. If we change this here, we should do this for all the other modules (like 'rest'), too. We could also implement a convention with directories without prefix but artifactId with prefix. But again, consistency (== convention) is king. 

But I'm ok with this, for now, will revisit the module names later as a whole.

syndesis-builder-image-generator/src/main/java/io/syndesis/image/Application.java
rhuss: Wouldn't a `mvn dependency:go-offline` be better anyway ?

PR #689, opened 2017-10-09
refactor(connectors): Bring back SQL stored connector

zregvart:
This re-enables the SQL stored procedure connector and the accompanying
connection by moving it from `demo-data.json` to `deployment.json`.


KurtStam:
@zregvart looks like the connectors version is set to 0.4-SNAPSHOT which makes a test fail.

zregvart:
> @zregvart looks like the connectors version is set to 0.4-SNAPSHOT which makes a test fail.

It might need a new connectors release, I think we have `latest` as the version, let me debug this a bit...

zregvart:
This is an issue in SQL stored connector, I've created https://github.com/syndesisio/connectors/pull/86. This PR should be updated when we release connectors.

KurtStam:
retest this please

KurtStam:
retest this please

KurtStam:
@zregvart I think this may need a rebase now that the connector project is released.

zregvart:
I've rebased this but I think the build will fail due to https://github.com/syndesisio/connectors/issues/88 so it will require a new version of connectors.

PR #690, opened 2017-10-10
fix: Allow snapshot deployments to the oss.sonatype.org mvn repos.

chirino:


KurtStam:
@chirino why do we want SNAPSHOTS?

jimmidyson:
I can't see any reason for this.

chirino:
I don't see any reason to not support it.

chirino:
At least with a snapshot, I can test out the script at: 
https://github.com/syndesisio/syndesis-s2i-image/blob/master/run.sh#L28

until a version gets released.

chirino:
retest this please

jimmidyson:
But we don't publish snapshots so we'd need to do that on each build too.

chirino:
I wouldn't do snapshot deploys as part of CI builds.  It's just nice to have an option for a developer to push snapshot build to the repo in case he needs to use it in another project and avoid waiting for the release to get cut.   This is definitely in the special case territory.

rhuss:
I think publishing of snapshots should never be an issue, only consuming it is a problem. I'm fine if people consume stuff temporarily without releasing snapshot dependencies. But that is checked during a release anyway. 

I just see this as a kind of 'scratch pad'. And CI shouldn't publish snapshots.

rhuss:
retest this please

PR #694, opened 2017-10-10
[WIP] feat: Change security configuration to header based auth from Keycloak

jimmidyson:
This PR drops Keycloak auth in favour of header based auth, requiring an oauth proxy (https://github.com/openshift/oauth-proxy) in production to delegate auth requests to OpenShift. Using the oauth proxy reduces our need to provide custom authentication strategy to delegate to OpenShift and perform proper SubjectAccessReviews - the oauth proxy does all that for us.

If required, we can add in other auth strategies for testing, etc. 

This requires updates to app templates before this is usable - I will submit a PR for that so we can merge at the same time. Please do not merge until this is available as that will also be part of my testing.


core/src/main/java/io/syndesis/core/KeyGenerator.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.syndesis.core;
 
+import net.iharder.Base64;
```
zregvart: There is an implementation of ordered base64 in `spring-security-crypto`[1], not sure if it makes much difference but would be a known dependency to us.

[1] https://docs.spring.io/autorepo/docs/spring-security/4.2.3.RELEASE/apidocs/org/springframework/security/crypto/codec/Base64.html#ORDERED

pom.xml
```diff
@@ -178,6 +177,8 @@
     <process-exec-maven-plugin.version>0.8</process-exec-maven-plugin.version>
     <swagger-maven-plugin.version>1.0.0</swagger-maven-plugin.version>
     <swagger2markup-maven-plugin.version>1.1.0</swagger2markup-maven-plugin.version>
+    <sulky-.version>8.1.1</sulky-.version>
```
zregvart: I don't see this being used, am I missing something?

zregvart:
I can see where the token is being verified, I know this is a WIP so this might be still on the TODO. Should we call out to some OpenShift API to check that the token is valid?

jimmidyson:
We don't need to validate the token, the oauth proxy does that auth for us. It's a trust issue, bind to localhost in the pod so only access to the REST API can be from the oauth proxy running in same pod and the only external accessible service.

core/src/main/java/io/syndesis/core/KeyGenerator.java
jimmidyson: That doesn't do what we want unfortunately... Just tried to migrate across and only private, internal methods do what we want.

pom.xml
jimmidyson: Removing

zregvart:
> We don't need to validate the token, the oauth proxy does that auth for us. It's a trust issue, bind to localhost in the pod so only access to the REST API can be from the oauth proxy running in same pod and the only external accessible service.

Awesome stuff! Really like the simplicity and separation of concerns here, we get everything we had with Keycloak in our UC 💯

jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
```diff
@@ -209,9 +209,9 @@ public void testPush() throws IOException {
     @Test
     public void testCreateKey() {
         String lastkey = jsondb.createKey();
-        for (int i = 0; i < 20000; i++) {
+        for (int i = 0; i < 200000; i++) {
```
rhuss: I know, its not related to this PR but since you touched this test, I believe this test can be removed completely as the same test is used in KeyGeneratorTest and `JsonDB.createKey()` just delegates to this method.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -129,7 +128,6 @@ public Integration create(@ConvertGroup(from = Default.class, to = AllValidation
             .createFrom(integration)
             .deployedRevisionId(revision.getVersion())
             .addRevision(revision)
-            .userId(Tokens.getUsername())
```
rhuss: This is currently an issue as the check for how many integrations are allowed is connected to the userName which is put as a label to the resource objects (see. e.g. [here](https://github.com/rhuss/ipaas-rest/blob/41b954a25817498ab68cb4351013ffc7d0b56bfe/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java#L115)). 

Despite the fact, I that the limitation should be global anyway (not per user), so this can be changed easily, I guess at least when we reach to the "User & Teams" use case we need a way to get to the subject of the current  http request (which might need to be added to the various domain objects for auditing purposes).

Is it easy to extract such a subject from Spring Security ? Maybe we can already add it here as user id ?

zregvart: Should be super easy you can either use the [static helpers](https://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/context/SecurityContextHolder.html#getContext--)

    SecurityContextHolder.getContext().getAuthentication()

Or you can inject it into a REST handler method I think there should be one for JAX-RS, in spring-web it's super simple, just add Authentication parameter to the method.

rhuss:
I guess we should merge this ASAP as the related template PR has been already merged ?

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
rhuss: @jimmidyson could change this line to use the Spring Security subject as the userId here ? As said, this is required currently for the integration limit feature.

jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
jimmidyson: Done.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
jimmidyson: Updated

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -151,7 +151,6 @@ public void update(String id, @ConvertGroup(from = Default.class, to = AllValida
         Integration updatedIntegration = new Integration.Builder()
             .createFrom(integration)
             .deployedRevisionId(existing.getDeployedRevisionId())
-            .userId(Tokens.getUsername())
```
rhuss: Guess its ok to remove this hear as the user has been taken over from the original integration.

rhuss:
retest this please

rhuss:
For me on minishift it works in principle, however the readiness checks with curl seem to fail constantly:

```
  58m		58m		1	{default-scheduler }							Normal		Scheduled	Successfully assigned syndesis-rest-1-0j1gb to localhost
  58m		58m		1	{kubelet localhost}	spec.initContainers{openshift-ca-pemtokeystore}	Normal		Pulled		Container image "docker.io/syndesis/pemtokeystore:v0.2.1" already present on machine
  58m		58m		1	{kubelet localhost}	spec.initContainers{openshift-ca-pemtokeystore}	Normal		Created		Created container
  58m		58m		1	{kubelet localhost}	spec.initContainers{openshift-ca-pemtokeystore}	Normal		Started		Started container
  58m		58m		1	{kubelet localhost}	spec.containers{syndesis-oauthproxy}		Normal		Pulled		Container image "docker.io/openshift/oauth-proxy:v1.0.0" already present on machine
  58m		58m		1	{kubelet localhost}	spec.containers{syndesis-oauthproxy}		Normal		Created		Created container
  58m		58m		1	{kubelet localhost}	spec.containers{syndesis-oauthproxy}		Normal		Started		Started container
  58m		55m		2	{kubelet localhost}	spec.containers{syndesis-rest}			Normal		Started		Started container
  58m		54m		5	{kubelet localhost}	spec.containers{syndesis-rest}			Warning		Unhealthy	Readiness probe failed: curl: (7) Failed connect to localhost:8181; Connection refused

  58m	51m	3	{kubelet localhost}	spec.containers{syndesis-rest}	Normal	Pulled		Container image "docker.io/syndesis/syndesis-rest:latest" already present on machine
  55m	51m	2	{kubelet localhost}	spec.containers{syndesis-rest}	Normal	Killing		Killing container with id docker://syndesis-rest:pod "syndesis-rest-1-0j1gb_myproject(bc8b1333-b0c4-11e7-a393-1a64ae70e8a2)" container "syndesis-rest" is unhealthy, it will be killed and re-created.
  58m	51m	3	{kubelet localhost}	spec.containers{syndesis-rest}	Normal	Created		Created container
  31m	28m	17	{kubelet localhost}	spec.containers{syndesis-rest}	Warning	BackOff		Back-off restarting failed container
  55m	8m	24	{kubelet localhost}	spec.containers{syndesis-rest}	Warning	Unhealthy	Liveness probe failed: curl: (22) The requested URL returned error: 404 Not Found

  31m	3m	72	{kubelet localhost}		Warning	FailedSync	Error syncing pod
```

This pod has been restarted already 12 times. Happens for syndesis-rest and syndesis-atlasmap, not for syndesis-verifier, though:

```
NAME                         READY     STATUS      RESTARTS   AGE
syndesis-atlasmap-2-deploy   0/1       Error       0          57m
syndesis-atlasmap-3-1vfxz    2/2       Running     10         45m
syndesis-db-1-591h7          1/1       Running     0          2h
syndesis-rest-1-0j1gb        2/2       Running     13         1h
syndesis-ui-1-s08s3          2/2       Running     0          1h
syndesis-verifier-1-rsrn8    1/1       Running     0          2h
todo-1-build                 0/1       Completed   0          2h
todo-1-w8rh6                 1/1       Running     0          2h
```

jimmidyson:
Which template are you using?

rhuss:
@jimmidyson syndesis-dev-rest. I just updated the liveness/readiness checks (see latest PR), this seems to work smoothely.

The first smoke test on our cluster fails though, because of resource limits:

```
Events:
  FirstSeen	LastSeen	Count	From				SubObjectPath	Type		Reason		Message
  ---------	--------	-----	----				-------------	--------	------		-------
  1m		1m		1	{replication-controller }			Warning		FailedCreate	Error creating: pods "syndesis-rest-3-" is forbidden: [maximum memory usage per Pod is 1Gi, but limit is 1178599424., maximum cpu usage per Pod is 2, but limit is 2195m.]
  1m		19s		13	{replication-controller }			Warning		FailedCreate	Error creating: pods "syndesis-rest-3-" is forbidden: [maximum cpu usage per Pod is 2, but limit is 2195m., maximum memory usage per Pod is 1Gi, but limit is 1178599424.]
```

Same for atlasmap. Unfortunately  I can' t continue today, but will continue tomorrow.

To test on the cluster is easy:

Just create a new project, install the dev-template locally, apply and then `oc edit` dc rest and ui to point to docker image which tag `tp2-smoke` (just pushed this to dockerhub).

see ya ....

rhuss:
retest this please

PR #707, opened 2017-10-16
refactor: use same Maven repositories thought

### LINKED ISSUE #664: Project generator should use the same Maven repositories as the the application
We have Maven repositories defined in the `application.yml` for Camel catalog, we should use the same repositories in the project generator for generating `pom.xml` of the integration project.

zregvart:
Adds MavenProperties that currently holds only Maven repositories that
the connector catalog, project generator and builder image generator
use.

Defines by default to use only Maven central.

Fixes #664

### DIFF
--- a/project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
+++ b/project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
@@ -44,6 +44,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import io.syndesis.connector.catalog.ConnectorCatalog;
 import io.syndesis.connector.catalog.ConnectorCatalogProperties;
+import io.syndesis.core.MavenProperties;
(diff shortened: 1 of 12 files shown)


project-generator/src/test/java/io/syndesis/project/converter/DefaultProjectGeneratorTest.java
```diff
@@ -73,7 +74,10 @@
 @RunWith(Parameterized.class)
 public class DefaultProjectGeneratorTest {
     private static final String CONNECTORS_VERSION = ResourceBundle.getBundle("test").getString("connectors.version");
-    private static final ConnectorCatalogProperties CATALOG_PROPERTIES = new ConnectorCatalogProperties();
+    private static MavenProperties mavenProperties = new MavenProperties(map("maven.central", "https://repo1.maven.org/maven2",
+        "redhat.ga", "https://maven.repository.redhat.com/ga",
```
rhuss: What part is using the redhat repo ?

zregvart: Just the test to make sure that if we need a repository it would get added to the integration POM, that part was refactored from static initializer.

PR #712, opened 2017-10-17
feat: Add "who am I" user endpoint

jimmidyson:
See https://github.com/syndesisio/syndesis-project/issues/134#issuecomment-337051867


rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
```diff
@@ -15,30 +15,49 @@
...
     public Kind resourceKind() {
         return Kind.User;
     }
-}
 
+    @Path("~")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public User whoAmI() {
+        String token = String.valueOf(SecurityContextHolder.getContext().getAuthentication().getCredentials());
```
rhuss: Doesn't the authentication subject contains something like a user name, too ? (I ask because we use that it at some other place).

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
```diff
@@ -15,30 +15,49 @@
...
     }
 
-     @Override
+    @Override
     public Kind resourceKind() {
         return Kind.User;
     }
-}
 
+    @Path("~")
```
rhuss: What does this kind of path mean ?

chirino: Thats just a static path to "~".  I like it since ~ has a similar meaning on the CLI.

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
zregvart: there is a [getPrincipal()](https://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/Authentication.html#getPrincipal--) method that should return the username

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
zregvart: Showoffs

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
jimmidyson: This returns a User object containing extra info, eg full name, so we need to make a call to auth source. In time, we can expand this to include roles, etc if required.

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
jimmidyson: Copied from openshift... I like it because it's a special character that is highly unlikely to be in anyone's username and not too annoying to disallow for users that might want to create their own usernames.

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
rhuss: ok, just want to get sure that getPrincipal() also works at another place

rhuss:
retest this please

rhuss:
retest this please

iocanel:
retest this please

rhuss:
@iocanel @jimmidyson this PR needs a rebase on master as we fixed the fork-timeout issue there.

rhuss:
just did the rebase

rhuss:
@jimmidyson a real thing found by CI (dependency issue). Good end to make peace with today. Could you check this, please ?

rest/src/main/java/io/syndesis/rest/v1/handler/user/UserHandler.java
jimmidyson: Yes it does. This is only to retrieve further info about the user that is not available in the oauth callback/headers.

rhuss:
[merge]

PR #717, opened 2017-10-17
fix: builder image generator was a little broken.

chirino:
It would not startup due to missing commons logging lib.


chirino:
Jenkins env seems borked.  Manual merging.

Issue #718, opened 2017-10-17
Couldn't find the user of the integration


paoloantinori:
I think there might be some out of sync data with the default Integrations already available.

Each time I try to start one of them, for instance, the simple `Couldn't find the user of the integration`

I get this error in `syndesis-rest` logs:

```
2017-10-17 19:44:15.361 ERROR [-,,,] 1 --- [pool-3-thread-1] i.s.c.integration.IntegrationController  : Error while processing integration status for integration 3

java.lang.IllegalStateException: Couldn't find the user of the integration
	at io.syndesis.controllers.integration.online.ActivateHandler.lambda$countIntegrations$1(ActivateHandler.java:169) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at java.util.Optional.orElseThrow(Optional.java:290) ~[na:1.8.0_131]
	at io.syndesis.controllers.integration.online.ActivateHandler.countIntegrations(ActivateHandler.java:169) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.controllers.integration.online.ActivateHandler.execute(ActivateHandler.java:66) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at io.syndesis.controllers.integration.IntegrationController.lambda$callStatusChangeHandler$10(IntegrationController.java:175) ~[controllers-0.1-SNAPSHOT.jar!/:0.1-SNAPSHOT]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_131]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) ~[na:1.8.0_131]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) ~[na:1.8.0_131]
	at java.lang.Thread.run(Thread.java:748) ~[na:1.8.0_131]

```

The code at the indicate line is:

```java
        String username = integration.getUserId().orElseThrow(() -> new IllegalStateException("Couldn't find the user of the integration"));
```
https://github.com/syndesisio/syndesis-rest/blob/master/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java#L108

[...]

PR #726, opened 2017-10-18
Setting the UserId on creation of connections and integrations #718

KurtStam:


model/src/main/java/io/syndesis/model/integration/Integration.java
```diff
@@ -73,7 +73,7 @@ default Kind getKind() {
 
     Optional<String> getIntegrationTemplateId();
 
-    Optional<String> getUserId();
+    String getUserId();
```
jimmidyson: Doesn't this mean the UI will have to send the userId every time with the request?

KurtStam: Yes think so, which is why I asked for @gashcrumb's review to make sure this is ok.

zregvart: Should we trust the username provided by the UI? Can we add a default method so its populated from Spring Security if not set already?

KurtStam: Yeah I like the idea of setting it on the backend. If we do that we should not even have the UI worry about it.

jimmidyson: We can add that as a default value, but would require muddying dependencies by adding spring secuity as a dependency to model.

rhuss:
Could we please hold this off until TP2 is released ? Thanks !

KurtStam:
@rhuss Sure we can wait. 

@jimmidyson I changed the PR so it leaves userId optional but it gets set when connections and integrations are created using the UserPrincipal.

KurtStam:
retest this please

rest/src/main/java/io/syndesis/rest/v1/operations/Creator.java
```diff
@@ -30,7 +32,7 @@
     @POST
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes("application/json")
-    default T create(@NotNull @Valid T obj) {
+    default T create(@Context SecurityContext sec, @NotNull @Valid T obj) {
```
jimmidyson: Is this needed here?

jimmidyson: Oh scratch that, it's used in subclass. Hmm not exactly liking this, but can't see a way round it I guess.

KurtStam:
retest this please

Issue #619, opened 2017-09-19
event stream SSE connection closes after about a minute


gashcrumb:
I raised this issue a few days ago -> https://github.com/syndesisio/syndesis-ui/issues/967

and am working on adding client-side code to handle reconnecting.  Couple questions though:

1)  What happens to stale/old reservations?  Am hoping when I get disconnected I can just claim a new reservation and reconnect
2)  Is there a gateway timeout kicking in that's closing the connection on us here that we could tweak?  The connection is dropped consistently.

chirino:
1. You can just create a new reservation.  The only issue is that you will only get event from the point in time the reservation is claimed.  You might drop event between the disconnect and the claim getting established.

2. Yeah, might be.. might also be idle settings between the client and server not be exactly right.  Undertow did no let me configure those.  Need to recheck to see if a more recent version allows it.

PR #728, opened 2017-10-18
fix: #619 Set 25 second keep alive timers on both SSE and WebSocket c…

chirino:
…onnections since open shift likes to close out idle backend server connection.


chirino:
if someone comes back requesting a different value, lets re-evaluate if we want to make it configurable.

chirino:
retest this please

chirino:
Jenkins seems busted still.

rhuss:
Come on, this can't be our default working mode. There is a reason for testing. I agree that Jenkins is a big PITA with our setup, but we need another solution urgently. Manual merging can't be the new standard // @iocanel let's bring it on are revive our CI discussion in hindsight of the latest troubles. Going to continue on tomorrow ....

iocanel:
@rhuss: Agree. We need a solution ASAP. I am not sure if the particular issues (that we have the last couple of weeks) has anything to do with our setup. It's a regression in one of (kubernetes-plugin, durable-task-plugin) and I am working on it.

PR #743, opened 2017-10-23
Adding filestore with DB backend

### LINKED ISSUE #735: Tech extension : add filestore with initial db backend
Tech extension will be uploaded as binary blob (jar) we need to store on back-end side thus we need a filestore with pluggable storage. Initial storage implementation should be database based.

nicolaferraro:
First PR on syndesis :smile: 
This fixes #735.

It adds a Filestore that can be used to store binary artifacts such as JARs.
The interface allows to do operation on filesystem-like paths and can be adapted to store artifacts in whatever storage we want in the future.

Current implementation uses a DB table called `filestore` inside the postgres database with 2 columns (`path` and `data`).

All the required file operation should be already implemented (write, load, move, delete and temp file support). Others can be added as needed.

### DIFF
--- a/filestore/src/main/java/io/syndesis/filestore/FileStoreException.java
+++ b/filestore/src/main/java/io/syndesis/filestore/FileStoreException.java
@@ -1,0 +1,44 @@
+/**
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.filestore;
+
+/**
+ * These runtime exceptions are thrown by the methods of the FileStore
+ */
+public class FileStoreException extends RuntimeException {
+
+    public FileStoreException(String message) {
+        super(message);
+    }
+
+    public FileStoreException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public static RuntimeException launderThrowable(Throwable cause) {
+        return launderThrowable("An error has occurred.", cause);
+    }
+
+    public static RuntimeException launderThrowable(String message, Throwable cause) {
+        if (cause instanceof RuntimeException) {
+            return ((RuntimeException) cause);
+        } else if (cause instanceof Error) {
+            throw ((Error) cause);
+        } else {
+            throw new FileStoreException(message, cause);
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


nicolaferraro:
@KurtStam do I need to fix *all* codacy "errors" or it's there as suggestion?

rhuss:
We all love  codacy ;-) It's ok to fix only the reasonable findings, and just ignore the stupid ones (we laready have quite struct PMD and findbugs checks, so codacy adds some sugar on top of it).

filestore/src/main/java/io/syndesis/filestore/FileStoreException.java
```diff
@@ -0,0 +1,30 @@
...
+public class FileStoreException extends RuntimeException {
+
+    public FileStoreException(String message) {
+        super(message);
+    }
+
+    public FileStoreException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
```
KurtStam: Can you add a launderThrowable method and use it? Like https://github.com/syndesisio/syndesis-rest/blob/master/core/src/main/java/io/syndesis/core/SyndesisServerException.java#L43-L55

filestore/src/main/java/io/syndesis/filestore/impl/SqlFileStore.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Implementation of a {@code FileStore} backed by a SQL database.
+ */
+public class SqlFileStore implements FileStore {
+
+    enum DatabaseKind {
+        PostgreSQL, H2
```
KurtStam: Most of our DB testing is done on Derby, can we add support for that too? Not sure we will use H2, we may want to drop that to simplify maintenance.

filestore/src/main/java/io/syndesis/filestore/FileStoreException.java
nicolaferraro: I can add a launderThrowable but the filestore exception is always used to add context information (e.g. the path being written) when a low level exception occurs. Since DBI exceptions (the low level exceptions) are always runtime exceptions, the message would be discarded, the runtime exception rethrown as is.. and this exception would not be used..

KurtStam: Yes ignore my comment, you actually want it to be a runtime exception.

nicolaferraro:
@KurtStam I've implemented some changes, including testing on derby and `launderThrowable` where best applicable.

Unfortunately I discovered I was too optimistic on thinking that BLOB support was handled correctly by JDBI, but after testing with Derby and with a external Postgres instance, I've added some custom code to handle the specific details of each database.

It would be good if I can replicate in CI the tests that I did manually on postgres, because there's a lot of difference between the way Postgres, H2 and Derby manage BLOBs. Are such tests supported by current CI?

KurtStam:
Thanks @nicolaferraro, unfortunately only our e2e tests run with Postgres at the moment.

rhuss:
@nicolaferraro is this an intrusive PR ? Just asking because we are still on the TP2 train. BTW, we wanted to introduce feature toggle to switch on/off things while we are going. this is just starting, though.

nicolaferraro:
I've added a `filestore.enabled` flag for this, and it defaults to `false`, so the filestore is not created (call it a basic version of feature toggle).

Just checked on syndesis to be sure.

rhuss:
@nicolaferraro cool, thanks. was just asking ;-)

Wonder though whether we should group feature toggles, so to have a top-leve `toggles` element with sub-elements like `filestore`. So excactly the other way round. So its easier to see what feature s toggles are available.

PR #744, opened 2017-10-23
refactor: provide default AuthenticationManager

zregvart:
By providing default AuthenticationManager bean to Spring application
context Spring Security auto configuration
(o.s.b.a.s.AuthenticationManagerConfiguration) is not triggered, and no
in-memory user details service will be created. This prevents the
message:

    Using default security password: ...

from being displayed at startup.


rhuss:
@zregvart Does this affect TP2 ? I ask because I do need to make a new build ...

PR #746, opened 2017-10-23
Refreshing data in the db

KurtStam:


rhuss:
@kstam would this affect TP2 ? I need to make a new build ...

KurtStam:
@rhuss no it should not, we don't load the demo-data.json in TP2. The deployment.json did not change so reloading that should have no effect. (Unless there is other data preloaded that I don't know about)

Issue #747, opened 2017-10-23
OpenShift error when creating build configuration


rhuss:
```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://openshift.default.svc/apis/build.openshift.io/v1/namespaces/proj7630/builds?labelSelector=openshift.io/build-config.name%3Dtwsf&fieldSelector=status%3DRunning. Message: No field label conversion function found for version: build.openshift.io/v1. Received status: Status(apiVersion=v1, code=400, details=null, kind=Status, message=No field label conversion function found for version: build.openshift.io/v1, metadata=ListMeta(resourceVersion=null, selfLink=null, additionalProperties={}), reason=BadRequest, status=Failure, additionalProperties={}).
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:470) ~[kubernetes-client-2.6.0-000015.fuse-000002.jar!/:na]
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:409) ~[kubernetes-client-2.6.0-000015.fuse-000002.jar!/:na]
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379) ~[kubernetes-client-2.6.0-000015.fuse-000002.jar!/:na]
```

the corresponding code looks like

```java
    @Override
    public boolean isBuildStarted(String name) {
        String sName = Names.sanitize(name);
        return !openShiftClient.builds()
                               .withLabel("openshift.io/build-config.name", sName)
                               .withField("status", "Running")
                               .list().getItems().isEmpty();
    }

```

and has not changed since 6 weeks.

@iocanel any ideas what this could be ?

iocanel:
@rhuss: I haven't seen that before. With a quick search I found:

https://github.com/openshift/jenkins-sync-plugin/issues/173 and some other related issues. 

Where do u get this?

rhuss:
I get this when I create when checking for a running build for the integration runtime on our ignite cluster (3.6.) with the images used from the pipeline builds. I never have seen this before, too. I'm a bit afraid that this is because our upstream images use a very old kubernetes-client (2.2.13), whereas the pipeline builds have 2.6.0. For 2.2.13 I never have seen this issue.

rhuss:
[...]

PR #748, opened 2017-10-24
chore: Update to kubernetes-client 2.4.1

rhuss:
That's the last one not triggering https://github.com/openshift/jenkins-sync-plugin/issues/173 and https://github.com/syndesisio/syndesis-rest/issues/747


zregvart:
@rhuss there is an issue found by the dependency check:

```
[ERROR] Found a problem with the direct dependency io.fabric8:kubernetes-model of the current project
  Expected version is 1.0.67
  Resolved version is 1.0.67
  Version 1.0.74 was expected by artifacts: io.fabric8:kubernetes-client, io.fabric8:openshift-client
```

PR #754, opened 2017-10-25
refactor: sort Maven repositories

zregvart:
This sorts configured Maven repositories by their keys as a simple
priority mechanism. Now if we specify environment variable like:

    MAVEN_REPOSITORIES_00_some_where=http://over.the.rainbow

The order of generated `pom.xml` of the integration pod and order by
which Grape resolves artifacts of Camel Connectors should be affected.


KurtStam:
retest this please

rhuss:
retest this please

PR #757, opened 2017-10-26
feat(connector-template): adds connector template support and Swagger Connector generator

### LINKED ISSUE #756: Implement Meta-Connectors and Swagger support
Meta-Connectors are templates for Connectors, and from their parameters supporting implementations derive Connectors. Implement Meta-Connector entity and a framework for processing parameters and generating Connectors with initial support for Swagger specifications.

zregvart:
Connector templates can be considered blueprints for creating
Connectors. Given a connector template a Connector generator
implementation generates a Connector.

This adds a SwaggerConnectorGenerator to sculpt the Connector with
actions, shapes and properties from Swagger specification.

Fixes #756


rhuss:
A good idea would be to align API changes with @nicolaferraro , who introduced `v1beta` kind of endpoints in https://github.com/syndesisio/syndesis-rest/pull/758

zregvart:
I wonder of anyone has a better suited name instead of `meta`? The code already uses this _sculptor_ metaphor would that be a better module name?

cc @syndesisio/backend

zregvart:
@rhuss I'm open for new name suggestions. I don't think this should be behind a `v1beta`, I'm not entirely sure as to what `v1beta` signifies, we have only one client and this doesn't modify existing endpoints (adds a new one).
I'll table the feature toggle until we decide how to handle feature toggles, I'm afraid that by adding it now I'll work on something I need to rework later.

KurtStam:
`ConnectorGenerators` and `ConnectorTemplates`

KurtStam:
@zregvart  Since we seem to be waiting for a rename is this still in WIP mode? Also should this be sprint 20 now?

zregvart:
I'll go with `connector-generator` suits the `project-generator` naming we have already.

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+        final String specification = configuredProperties.get("specification");
+
+        if (specification == null) {
+            throw new IllegalStateException(
+                "Configured properties of the given Connector template does not include `specification` property");
+        }
+
+        final Connector baseConnector = baseConnectorFrom(connectorTemplate, template);
+
+        final Connector connector = new Connector.Builder()//
```
jimmidyson: nit: more extra `//`

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+            .build();
+
+        return configureConnector(connectorTemplate, connector, specification);
+    }
+
+    /* default */ static void addGlobalParameters(final Connector.Builder builder, final Swagger swagger) {
+        final Map<String, Parameter> globalParameters = swagger.getParameters();
+        if (globalParameters != null) {
+            globalParameters.forEach((name, parameter) -> {
+                final Optional<ConfigurationProperty> property = createPropertyFromParameter(parameter);
```
jimmidyson: Can be refactored to simply:

```java
createPropertyFromParameter(parameter).ifPresent((property) -> {builder.putProperty(name, property)});
```

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+
+            for (final Entry<HttpMethod, Operation> entry : operationMap.entrySet()) {
+                final Operation operation = entry.getValue();
+                if (operation.getOperationId() == null) {
+                    operation.operationId("operation-" + idx++);
+                }
+
+                final ActionDefinition actionDefinition = createActionDefinition(specification, operation);
+
+                final Action action = new Action.Builder()//
```
jimmidyson: nit yet more `//`

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /* default */ static ActionDefinition createActionDefinition(final String specification,
+        final Operation operation) {
+        final ActionDefinition.Builder actionDefinition = new ActionDefinition.Builder();
+
+        final Optional<BodyParameter> maybeRequestBody = operation.getParameters().stream()
+            .filter(p -> p instanceof BodyParameter && ((BodyParameter) p).getSchema() != null)
+            .map(BodyParameter.class::cast).findFirst();
+        if (maybeRequestBody.isPresent()) {
```
jimmidyson: Could be rewritten as (something like anyway!):

```java
actionDefinition.inputDataShape(
    maybeRequestBody
        .flatMap((requestBody) -> createShapeFromModel(specification, requestBody.getSchema()))
        .orElse(DATA_SHAPE_NONE)
);
```

Forgive me for playing with `Optional`s - not 100% sure which is more readable tbh...

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+            final BodyParameter requestBody = maybeRequestBody.get();
+            actionDefinition.inputDataShape(createShapeFromModel(specification, requestBody.getSchema()));
+        } else {
+            actionDefinition.inputDataShape(DATA_SHAPE_NONE);
+        }
+
+        final Optional<Response> maybeResponse = operation.getResponses().values().stream()
```
jimmidyson: Again could be rewritten with `Optional.flatMap` if you feel it.

connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+        for (final Parameter parameter : operation.getParameters()) {
+            final Optional<ConfigurationProperty> property = createPropertyFromParameter(parameter);
+
+            if (property.isPresent()) {
+                step.putProperty(parameter.getName(), property.get());
+            }
+        }
+
+        step.putProperty("operationId",
+            new ConfigurationProperty.Builder()//
```
jimmidyson: nit: more extra `//`

pom.xml
```diff
@@ -563,6 +582,12 @@
       </dependency>
 
       <dependency>
+        <groupId>com.google.guava</groupId>
```
jimmidyson: What's this new dependency for?

pom.xml
```diff
@@ -575,6 +600,12 @@
       </dependency>
 
       <dependency>
+        <groupId>com.google.errorprone</groupId>
```
jimmidyson: Where is this dependency being used?

pom.xml
```diff
@@ -604,6 +635,39 @@
...
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+
+      <dependency>
+        <groupId>io.swagger</groupId>
+        <artifactId>swagger-models</artifactId>
+        <version>1.5.16</version>
```
jimmidyson: Use `${swagger.version}`

connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+import io.syndesis.core.Names;
+import io.syndesis.model.connection.Connector;
+import io.syndesis.model.connection.ConnectorTemplate;
+
+public interface ConnectorGenerator {
+
+    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate, final Connector template) {
+        final Set<String> properties = connectorTemplate.getProperties().keySet();
+
+        final Map<String, String> configuredProperties = template.getConfiguredProperties()//
```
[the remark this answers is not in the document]
zregvart: I use Eclipse IDE, this is a sad way to have the formatter not join those lines together, and I auto-format on Save. It's a sad world I live in.

connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import io.syndesis.core.Names;
+import io.syndesis.model.connection.Connector;
+import io.syndesis.model.connection.ConnectorTemplate;
+
+public interface ConnectorGenerator {
+
+    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate, final Connector template) {
```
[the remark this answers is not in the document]
zregvart: I agree it is a bit confusing and should be polished, it comes from the POC as it was easiest to use the `Connector` to hold the properties that the user specifies (icon, name, description and specification), so this remained from that. For the production version I think we need another DTO here.

`ConnectorTemplate` is fixed, defined in the database, holds properties that can be given (here in the `Connector`) and the resulting `Connector` contains configuredProperties given in the `Connector` defined by `properties` of the `ConnectorTemplate`, also property definitions from `ConnectorTemplate` in `connectorProperties`.

pom.xml
zregvart: Aaah, had to remove it to remember, `io.swagger:swagger-jaxrs` requires Guava 20.0, let me see if I can remove `swagger-jaxrs` instead.

This is the original error:
```
[ERROR] Found a problem with the direct dependency com.google.guava:guava of the current project
  Expected version is 18.0
  Resolved version is 18.0
  Version 20.0 was expected by artifact: io.swagger:swagger-jaxrs
```

zregvart: Yeah, can't do that we're using it in the `rest` module (`io.syndesis.rest.v1.SwaggerConfiguration`)

pom.xml
zregvart: Guava brings it in, and once it's in Immutables adds it to the generated code, if we don't declare it we get:

```
[INFO] --- maven-dependency-plugin:3.0.1:analyze-only (default) @ rest ---
[WARNING] Used undeclared dependencies found:
[WARNING]    com.google.errorprone:error_prone_annotations:jar:2.0.18:compile
```

jimmidyson:
@zregvart If you're happy to remove WIP label then this is good to merge, with exception of dodgy CI infra (platform issue).

jimmidyson:
PR build seems to be hanging after image push - @iocanel?

iocanel:
@jimmidyson, @rhuss: This is a a bug in our latest openshift cluster:https://github.com/openshift/origin/issues/17294.

What I propose is to make the pipelines ignore it (as the actual build is successful). Ideally we can have a verification process that ensures that image was actually build.

PR #758, opened 2017-10-26
Add API to manage tech extensions

nicolaferraro:
This adds a API to manage tech extensions.

Supported operations are:
- POST (this accepts a binary artifact and returns the JSON metadata of the extension)
- LIST
- GET
- DELETE

API are published under "/api/v1beta1". Beta api are disabled by default. They can be enabled with toggles (env var on Deployment):
```
FEATURES_API_V1BETA1_ENABLED=true
FEATURES_FILESTORE_ENABLED=true
```

I've put a toggle also for the "v1" api, but it's enabled by default.

When a binary artifact is uploaded, it's stored in "/extensions/[id]" into the filestore (same id of the entity in the jsondb). The artifact can be retrieved later (not implemented yet) by ID or GAV. A retrieval by GAV will require first a lookup into the jsondb.

Fixed some issues found with the filestore now that it's integrated in the loop (initialization on a dirty db was not working, added auto-detection of DB type, fixed PG connection management).

Added also IT tests (and done manual testing on Openshift).

### DIFF
--- a/dao/src/main/java/io/syndesis/dao/ExtensionDao.java
+++ b/dao/src/main/java/io/syndesis/dao/ExtensionDao.java
@@ -13,19 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.syndesis.rest.v1;
+package io.syndesis.dao;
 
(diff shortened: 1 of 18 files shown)


nicolaferraro:
np, closing this and opening separate PRs

PR #768, opened 2017-11-01
feat: Add integration export and import endpoints.

### LINKED ISSUE #764: Add integration export and import endpoints.
Need new endpoints:
* `GET /api/v1/integrations/{id}/export.zip` should download a .zip file which holds all the data need to re-import the given integration into a different environment.

* `POST /api/v1/integration-support/import` and posting it the exported zip should create/update the integration in the new env.

chirino:
Fixes #764

### DIFF
--- a/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
+++ b/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
@@ -16,10 +16,16 @@
 package io.syndesis.rest.v1.handler.integration;
 
 import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiParam;
+import io.syndesis.controllers.EncryptionComponent;
+import io.syndesis.core.Json;
+import io.syndesis.dao.init.ModelData;
 import io.syndesis.dao.manager.DataManager;
 import io.syndesis.inspector.Inspectors;
 import io.syndesis.model.Kind;
 import io.syndesis.model.ListResult;
+import io.syndesis.model.connection.Connection;
+import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.DataShape;
 import io.syndesis.model.filter.FilterOptions;
 import io.syndesis.model.filter.Op;
@@ -27,6 +33,7 @@ import io.syndesis.model.integration.Integration;
 import io.syndesis.model.integration.Integration.Status;
 import io.syndesis.model.integration.IntegrationRevision;
 import io.syndesis.model.integration.IntegrationRevisionState;
+import io.syndesis.model.integration.Step;
 import io.syndesis.model.validation.AllValidations;
 import io.syndesis.rest.util.PaginationFilter;
 import io.syndesis.rest.util.ReflectiveSorter;
@@ -39,28 +46,37 @@ import io.syndesis.rest.v1.operations.PaginationOptionsFromQueryParams;
 import io.syndesis.rest.v1.operations.SortOptionsFromQueryParams;
 import io.syndesis.rest.v1.operations.Updater;
 import io.syndesis.rest.v1.operations.Validating;
-import io.syndesis.controllers.EncryptionComponent;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Component;
 
 import javax.persistence.EntityNotFoundException;
 import javax.validation.Validator;
+import javax.validation.constraints.NotNull;
 import javax.validation.groups.ConvertGroup;
 import javax.validation.groups.Default;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.SecurityContext;
+import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
+import java.io.IOException;
 import java.math.BigInteger;
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import static io.syndesis.rest.v1.handler.integration.IntegrationSupportHandler.EXPORT_MODEL_FILE_NAME;
+
 
 @Path("/integrations")
 @Api(value = "integrations")
@@ -107,6 +123,44 @@ public class IntegrationHandler extends BaseHandler
         }
 
         return integration;
+    }
+
+
+    @GET
+    @Path("/{id}/export.zip")
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public StreamingOutput export(@NotNull @PathParam("id") @ApiParam(required = true) String id) throws IOException {
+        ArrayList<ModelData> models = new ArrayList<>();
+
+        Integration integration = this.get(id);
+        models.add(new ModelData(Kind.Integration, integration));
+
+        for (Step step : integration.getSteps()) {
+            Optional<Connection> c = step.getConnection();
+            if( c.isPresent() ) {
+                Connection connection = c.get();
+                models.add(new ModelData(Kind.Connection, connection));
+                Connector connector = getDataManager().fetch(Connector.class, connection.getConnectorId().get());
+                if( connector != null ) {
+                    models.add(new ModelData(Kind.Connector, connector));
+                }
+            }
+        }
+
+        return out -> {
+            try (ZipOutputStream tos = new ZipOutputStream(out) ) {
+                addEntry(tos, EXPORT_MODEL_FILE_NAME, Json.mapper().writeValueAsBytes(models));
+                // Eventually we might need to add things like tech extensions too..
+            }
+        };
+    }
+
+    private void addEntry(ZipOutputStream os, String path, byte[] content) throws IOException {
+        ZipEntry entry = new ZipEntry(path);
+        entry.setSize(content.length);
+        os.putNextEntry(entry);
+        os.write(content);
+        os.closeEntry();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


KurtStam:
Don't we need a check on the version of the model/api?

chirino:
@KurtStam I think that's something that we can add later. We generally need something like that for all our entities and perhaps we can leverage that here too.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -46,4 +63,83 @@ public IntegrationSupportHandler( ProjectGenerator projectConverter) {
...
+            while (true) {
+                ZipEntry entry = zis.getNextEntry();
+                if( entry == null ) {
+                    break;
+                }
+                if (EXPORT_MODEL_FILE_NAME.equals(entry.getName())) {
+                    ModelData[] models = Json.mapper().readValue(new FilterInputStream(zis) {
+                        @Override
+                        public void close() throws IOException {
+                            // We want to avoid closing zis
```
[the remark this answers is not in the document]
chirino: we don't want to close the zip stream since we wont be able to read from it again when we get to the next zip entry.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -46,4 +63,83 @@ public IntegrationSupportHandler( ProjectGenerator projectConverter) {
...
+                        }
+                    }, ModelData[].class);
+
+                    imported += importModels(models);
+                }
+                zis.closeEntry();
+            }
+            if (imported==0) {
+                LOG.info("Could not import integration: No integration data model found.");
+                return Response.status(Response.Status.BAD_REQUEST).build();
```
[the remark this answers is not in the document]
chirino: We need to think about what kind of error messages would actually help a citizen integrator before we go to deep down this route.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -46,4 +63,86 @@ public IntegrationSupportHandler( ProjectGenerator projectConverter) {
...
+                LOG.info("Could not import integration: " + e, e);
+            }
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+    }
+
+    public int importModels(ModelData... models) throws IOException {
+        int count = 0;
+        for (ModelData model : models) {
+            switch (model.getKind()) {
```
zregvart: Could this be simpler? We can get `modelClass` from `Kind` and could have a constant `EnumSet<Kind> ALLOWED_IMPORTS`?

zregvart: Aah I see Connections/Connectors are handled differently

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -108,6 +121,44 @@ public Integration get(String id) {
...
+                Connector connector = getDataManager().fetch(Connector.class, connection.getConnectorId().get());
+                if( connector != null ) {
+                    models.add(new ModelData(Kind.Connector, connector));
+                }
+            }
+        }
+
+        return out -> {
+            try (ZipOutputStream tos = new ZipOutputStream(out) ) {
+                addEntry(tos, EXPORT_MODEL_FILE_NAME, Json.mapper().writeValueAsBytes(models));
```
zregvart: Would a ZIP file with multiple entries make more sense? the file names could be sanitized names of Integrations/Connectors/Connections?

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -108,6 +121,44 @@ public Integration get(String id) {
...
+        Integration integration = this.get(id);
+        models.add(new ModelData(Kind.Integration, integration));
+
+        for (Step step : integration.getSteps()) {
+            Optional<Connection> c = step.getConnection();
+            if( c.isPresent() ) {
+                Connection connection = c.get();
+                models.add(new ModelData(Kind.Connection, connection));
+                Connector connector = getDataManager().fetch(Connector.class, connection.getConnectorId().get());
+                if( connector != null ) {
```
rhuss: In which case can the connector be null ? Why it is an optional on connection anyway ? (so you should check before calling get())

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -108,6 +121,44 @@ public Integration get(String id) {
...
+                    models.add(new ModelData(Kind.Connector, connector));
```
rhuss: Shouldn't you add a connector only if not already adedd ? (e.g. when you use the same connector but different actions for incoming and outcoming)

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
rhuss: For the moment I don't expect many people looking into the zip so its somewhat opaque. 

What I miss though is some sort of versioning, which allows for evolving the export/import format over time.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -108,6 +121,44 @@ public Integration get(String id) {
...
+        ArrayList<ModelData> models = new ArrayList<>();
+
+        Integration integration = this.get(id);
+        models.add(new ModelData(Kind.Integration, integration));
+
+        for (Step step : integration.getSteps()) {
+            Optional<Connection> c = step.getConnection();
+            if( c.isPresent() ) {
+                Connection connection = c.get();
+                models.add(new ModelData(Kind.Connection, connection));
```
rhuss: Shouldnt a connection be added only if not already added ?

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -46,4 +63,86 @@ public IntegrationSupportHandler( ProjectGenerator projectConverter) {
...
+                        dataManager.update(integration);
+                    }
+                    count ++;
+                    break;
+
+                case Connection:
+
+                    // We only create connections, never update.
+                    Connection connection = (Connection) model.getData();
+                    if (dataManager.fetch(Connection.class, connection.getId().get()) == null) {
```
rhuss: Is the id() some kind of symbolic name or a real (universal ?) unique id ? How it is is guaranteed that there is no clash with an already existing id (not coming from an import) ? Would it make sense to add some kind of unique prefix / namespace to imported connections to indicate their source ?

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -108,6 +121,44 @@ public Integration get(String id) {
         return integration;
     }
 
+
+    @GET
+    @Path("/{id}/export.zip")
```
rhuss: as the name used here is also used for storing the file, I would chose a more expressive name like 'ignite-integration-export.zip' (or 'syndesis-integraiton-export.zip').

Or, alternatively and probably better, set the Http-Header for `Content-Type: application/octet-stream` and `Content-Disposition: attachment; filename="syndesis-integration-export.zip"` (with customization for fuse-ignite in production)

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: I agree, but I think we also need this in the DB model for the same reasons.  So I think once we put that in place there, this export/import thing will likely pick it up for free.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: It's not really used to control the file name.  The ui module is in control of the the file name.  see: https://github.com/syndesisio/syndesis-ui/blob/master/src/app/integrations/detail-page/detail.component.ts#L413

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: I considered this, but there is no harm in exporting it multiple times and it should be a very unlikely scenario.  Keeping code simple wins in this case.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: It would be an odd case where the DB is not consistent.  Generally I think all this code could use more null/optional checking.  But this is more of a POC anyways.  (@dsimansk: would like these edge cases to be found by tests)

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: Once again there is no harm in exporting the same thing multiple times.  It only gets imported once.

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
chirino: Right!

rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
chirino: Our id's are GUIDs, there should be no clashes so I think this is safe.
matching on the name of the object would be the other option, but is less safe since you could have conflicts, but might make more sense since folks are in control of that more.

rhuss:
I'm wrong WRT the connector ids, as they are already name-like (not a random id as for the integration).

rhuss:
@chirino There's a conflict we you need to resolve.

chirino:
rebased and squashed.

PR #770, opened 2017-11-02
feat: Encrypt secrets before storing them in the DB.

### LINKED ISSUE #373: REST apis must not send credentials back to the client

### LINKED ISSUE #764: Add integration export and import endpoints.

chirino:
fixes #373

### DIFF
--- a/controllers/pom.xml
+++ b/controllers/pom.xml
@@ -69,6 +69,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-core</artifactId>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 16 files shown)


chirino:
It does require the template change.  If this update is rolled out /w the template, it is backward compatible at least from the DB perspective.

PR #771, opened 2017-11-02
fix: set the username as the label when creatin...

### LINKED ISSUE #769: Integration count limit is not enforced in TP2
It seems that our limits are not applied for TP2 on production cluster. Therefor I'm able to deploy 2 Integrations at once. Fortunately the current quota(after Keycloak removal) is gracious enough to accomodate those deployments.

zregvart:
...g integration deployment

Username that we use to count the number of deployments was set as
annotation and when counting expected as label. This sets the username
as label when creating integration deployment.

Fixes #769


dsimansk:
I'd prefer label with prefix like `syndesis.io`. I might be too picky though :)
 
https://github.com/syndesisio/syndesis-rest/blob/bfea32ff70468911d4b7ebc3c8b29d934f8e7d6d/openshift/src/main/java/io/syndesis/openshift/OpenShiftService.java#L28-L29

PR #772, opened 2017-11-02
Also encrypt the sampldb connection password.

### LINKED ISSUE #373: REST apis must not send credentials back to the client

chirino:
fixes #373

### DIFF
--- a/controllers/pom.xml
+++ b/controllers/pom.xml
@@ -74,11 +74,6 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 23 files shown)


chirino:
rebased and squashed.

KurtStam:
@chirino Looks there is still a conflict.

PR #775, opened 2017-11-03
 feat(credential): generic OAuth2 support

### LINKED ISSUE #727: Add support for generic OAuth
We need to support generic OAuth providers for syndesisio/syndesis-project#119

zregvart:
Given enough parameters OAuth2 support in credentials can be generic
enough, this adds support for using OAuth2 provider for any connector
that declares that it supports it by tagging a property with
`authentication-type` tag and having set that property value to
`oauth2`.

Other properties that are used also need to be tagged:
 - `oauth-access-token` - property that will hold the value of the OAuth
   token
 - `oauth-access-token-url` - URL for the auth code to token exchange
 - `oauth-authentication-url` - URL for the OAuth authentication
   endpoint
 - `oauth-authorization-url` - URL for the OAuth authorization endpoint
 - `oauth-authorize-using-parameters` - toggle true/false if using OAuth
   authorization via query parameters instead of basic-auth header
 - `oauth-token-strategy` - used with RestTemplate (not used by Syndesis
   currently)

For details about this options consult Spring Social project.

Fixes #727


zregvart:
Moved to https://github.com/syndesisio/syndesis/pull/396

PR #781, opened 2017-11-06
Fixes #141: Add validation to tech extension API

nicolaferraro:
This adds services to validate the tech extension in order to detect blocking/non-blocking errors (the HTTP status code determines the kind of errors). Both kind of alerts are computed using bean validation. I've used the `NonBlockingValidation` group for checking non-blocking errors.


KurtStam:
retest this please

Issue #751, opened 2017-10-24
Add new 'pattern' field to the connector 


KurtStam:
The Camel connector has a 'pattern' field we should capture. This field can contain 'To' and 'From' (and maybe 'anywhere' - have to look that up, if so then we'd need to add it to Action as well). And the UI should filter by it when displaying possible connectors/connections/actions. A 'Start' should only from 'From' patterns.

gashcrumb:
So the semantics for this require some thought:

* `From` shouldn't be in the last step
* `To` shouldn't be in the first step
* `Anywhere` obviously can always be used if we have this possibility too

Question though:

* Can a `From` be used in a middle step, or is it only for the first step?
* Likewise for a `To`, can it show up in the middle of an integration?

I think the answer for either above is probably yes, these could show up in the middle of an integration, but I could be wrong there :-)

KurtStam:
Pretty sure the `From` can be in the first step only. `to` can occur in multiple times and can be in the middle. https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/6.3/html/apache_camel_development_guide/basicprinciples

rhuss:
@KurtStam is you say that this is a field on a Camel connector, then you mean, that this can be reused for a Syndsis _Action_, right ? (camel-connector == Syndesis Action).

So a Connection should only be offered for a position only when _some_ of the associated actions has a matching pattern. And also when the actions are listed, only the actions which are valid for this position should be shown.

BTW, for me 'pattern' is a completely misleading name. Would like 'direction' much better. 

Or something like 'kind' and then with the values `Source`, `Sink`, `SourceSink` ....

gashcrumb:
+1 for `Source`, `Sink` and `SourceSink`, that's much easier to understand.  The data shapes should correlate to this, I guess one option is to use the presense of data shapes to figure this out too:

* `Source` -> only an output data shape
* `Sink` -> only an input data shape
* `SourceSink` -> both an output and an input data shape

This means then that a `Source` can only be used at the beginning of an integration and a `Sink` can only go at the end.

KurtStam:
@rhuss yes synthesis action would be where we would add it. Camel calls it `pattern`. Since we don't expose this name to the enduser I'd prefer to keep it the same. 

@gashcrumb 
- `From` only in the first Step
- `To` can't be the first Step

[...]

PR #782, opened 2017-11-06
feat: Adding 'pattern' parameter to specify if the integration is a '…

KurtStam:
…from' or 'to' #751

### DIFF
--- a/model/src/main/java/io/syndesis/model/connection/Action.java
+++ b/model/src/main/java/io/syndesis/model/connection/Action.java
@@ -30,12 +30,15 @@ import io.syndesis.model.WithConfigurationProperties;
 import io.syndesis.model.WithId;
 import io.syndesis.model.WithName;
 import io.syndesis.model.WithTags;
+
 import org.immutables.value.Value;
 
 @Value.Immutable
 @JsonDeserialize(builder = Action.Builder.class)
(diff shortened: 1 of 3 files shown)


KurtStam:
@rhuss Do I need to make this param Optional so it's backwards compatible for TP2?

model/src/main/java/io/syndesis/model/connection/Action.java
```diff
@@ -30,13 +30,16 @@
...
 import io.syndesis.model.WithTags;
+
 import org.immutables.value.Value;
 
 @Value.Immutable
 @JsonDeserialize(builder = Action.Builder.class)
 @JsonIgnoreProperties(value = {"properties", "inputDataShape", "outputDataShape"}, allowGetters = true)
 public interface Action extends WithId<Action>, WithName, WithTags, WithConfigurationProperties, Serializable {
 
+    enum EI_Pattern { from, to }
```
zregvart: Why not just `Pattern` `EI_Pattern` is so un-Java like? Perhaps `Use`?

In Camel these values are capitalized I wonder if it would be easier to have them like that also. It would save us some translation when needed. (e.g. `From` and `To`).

KurtStam:
retest this please

KurtStam:
retest this please

PR #786, opened 2017-11-07
fix: Remove tomcat dependency.

chirino:


KurtStam:
retest this please

chirino:
@KurtStam the error is not intermittent.  The failure:
```
[INFO] --- dependency-management-maven-plugin:0.7:analyze (default) @ runtime ---
[WARNING] Exclusions must be removed for managed dependency org.springframework.boot:spring-boot-starter-jdbc:jar
```

Is due to a doggy rule in the dependency-management-maven-plugin in my opinion.  The BOM managed dependency is defined in the spring boot BOM which is released and we can't change.  Should we just disable this plugin from running.  Seems like it leads to more trouble than what it helps to solve.

PR #810, opened 2017-11-13
Tech extension: improve maven plugin

lburgazzoli:


lburgazzoli:
@syndesisio/backend can anyone review ?

syndesis-annotation-processor/src/main/java/io/syndesis/maven/annotation/processing/SyndesisExtensionActionProcessor.java
```diff
@@ -198,7 +217,8 @@ protected File obtainResourceFile(Element element) throws IOException {
 
         final String fileName = new StringBuilder()
             .append(classElement.getSimpleName().toString())
-            .append(element.getAnnotation(SyndesisExtensionAction.class).id())
+            .append('-')
+            .append(UUID.randomUUID().toString())
```
nicolaferraro: My understanding is that each compilation could generate a different file, unless you use `mvn clean ...` on every run...

lburgazzoli: yeah will address it soon

Issue #776, opened 2017-11-03
Tech extension : maven archetype 


lburgazzoli:
We should have a maven archetype to generate tech extension projects templates

PR #812, opened 2017-11-13
Tech extension : maven archetype #776

lburgazzoli:


lburgazzoli:
retest this please

KurtStam:
@lburgazzoli, am I right to assume that we want to maintain this here as we think this project will be the only consumer? If so where are we going to publish the documentation for it?

lburgazzoli:
Mh not sure, what would be the best location for archetypes ?

lburgazzoli:
retest this please

lburgazzoli:
@KurtStam @rhuss would it be ok to keep this in syndesis-rest till we move to monorepo ? then we can make archetype a standalone module

rhuss:
@lburgazzoli looks. Get ready for the move, fasten seatbelts and then when can move when we are over the bridge.

